PR #4, opened 2015-05-07
Python 3 support

### LINKED ISSUE #3: Python3 support
What's the state of Python3 support with `pytest-html`?

In `html/plugin.py` I noticed this:

```python
# Python 2.X and 3.X compatibility
if sys.version_info[0] < 3:
    from codecs import open
```

However python3 tests were removed in 30a37b7b4ddb06db36659dd7c22740030fe9ca8d and it doesn't seem to work - one reason seems to be the `html` package name conflicts with the [html module in the stdlib](https://docs.python.org/3/library/html.html):

```
Downloading/unpacking pytest-html
  Downloading pytest-html-1.0.tar.gz
  Running setup.py (path:/tmp/pip-build-v2gumy9z/pytest-html/setup.py) egg_info for package pytest-html
    Traceback (most recent call last):
      File "<string>", line 3, in <module>
      File "/home/florian/proj/qutebrowser/git/.tox/unittests/lib/python3.4/site-packages/setuptools/__init__.py", line 12, in <module>
        from setuptools.extension import Extension
      File "/home/florian/proj/qutebrowser/git/.tox/unittests/lib/python3.4/site-packages/setuptools/extension.py", line 7, in <module>
        from setuptools.dist import _get_unpatched
      File "/home/florian/proj/qutebrowser/git/.tox/unittests/lib/python3.4/site-packages/setuptools/dist.py", line 16, in <module>
        from setuptools.depends import Require
      File "/home/florian/proj/qutebrowser/git/.tox/unittests/lib/python3.4/site-packages/setuptools/depends.py", line 6, in <module>
        from setuptools import compat
      File "/home/florian/proj/qutebrowser/git/.tox/unittests/lib/python3.4/site-packages/setuptools/compat.py", line 44, in <module>
        from html.entities import name2codepoint
    ImportError: No module named 'html.entities'
```

Would a pull request for python3 support be welcome? Then I'd like to get this working - it'll mean the package needs to be renamed however, e.g. to `pytest_html`.

Also, would it be okay to introduce [six](https://pypi.python.org/pypi/six) as a dependency if it makes things easier?

The-Compiler:
Indeed, `six` was not needed and it was easier than I expected to port it :smile:.

The tests pass locally with 2.7 and 3.4, so let's see what Travis says. It also works fine with the test suite of [my application](https://github.com/The-Compiler/qutebrowser).

After this is merged I'd appreciate a new release on PyPI (unless there's some reason it shouldn't be released yet of course), so I can easily use it for my tests.

Closes #3.


The-Compiler:
Okay, it seems for 3.2, `u"foo"` is a `SyntaxError`... There are essentially three solutions for this:

* Drop support for 3.2 - I'm not sure how widespread 3.2 is nowadays, but I know many projects don't support <= 3.2.
* Use `from __future__ import unicode_literals` ([details](http://python-future.org/unicode_literals.html)), remove the `u` prefixes, and add `b` prefixes as needed
* Use the `u("foo")` function from six:

```python
if PY3:
    def u(s):
        return s
else:
    # Workaround for standalone backslash
    def u(s):
        return unicode(s.replace(r'\\', r'\\\\'), "unicode_escape")
```

What do you think?

davehunt:
Yuck! I think I'm favouring option 2 (Use from `__future__ import unicode_literals`, remove the u prefixes, and add b prefixes as needed). Thanks for the pull request! :)

The-Compiler:
I found a much better solution for this case - just drop the `u` prefix (which was only used in one place), because it's an ascii-only string anyways and will be implicitely decoded to utf-8 because of `codecs.open(..., encoding='utf-8')`.

davehunt:
Great - I'll merge and push a new release to PyPI! Thanks!

davehunt:
Done: https://pypi.python.org/pypi/pytest-html/1.1

The-Compiler:
Awesome, thanks!

In case you're interested, I've now successfully integrated it into [my buildbot](http://qutebrowser.org:8010/waterfall) :smile:

davehunt:
That's awesome! :+1:

PR #6, opened 2015-05-19
Improve sorting

The-Compiler:
This sorts tests by the results column by default, and defines a custom sorting which shows the tests roughly in order of importance so failed/skipped tests stand out from (hopefully many) passed ones.

Note I didn't write much javascript before that, so a thorough review is appreciated - it seems to work fine for me though.

A new release after this would be appreciated :wink:


davehunt:
Nice enhancement, thanks! I've released a new version to PyPI.

PR #19, opened 2015-09-08
Preserve environment details when using pytest-xdist

davehunt:


davehunt:
Merged in https://github.com/davehunt/pytest-html/commit/2cdd063b4ca0d1b6f883d007c25324da15afdd3b

PR #20, opened 2015-09-16
added stdout/stderr captures to log

rpuntaie:
https://github.com/davehunt/pytest-html/issues/8

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -119,22 +119,27 @@ class HTMLReport(object):
                     target='_blank'))
                 links_html.append(' ')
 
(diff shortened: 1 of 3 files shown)


davehunt:
Thanks for the patch @rpuntaie. I took a slightly different approach and fixed this in https://github.com/davehunt/pytest-html/commit/49535899d75c282529bca3ec7b48530ce3e63839. I'm not sure about the value of the index in the HTML report though. Perhaps date/timestamps would make more sense if you want to sort the tests by execution?

rpuntaie:
A possibility to sort based on execution order would be fine. Since date is present in the report, a date stamp per entry might be too much redundancy. It is a viable solution, though.

Issue #27, opened 2016-01-25
pytest-html is trying to encode my data inappropriately


birdsarah:
I'm using:
* pytest-selenium 1.1
* pytest-html 1.7
* selenium 2.49.2

I get the following error on a test fail:

````python
INTERNALERROR>     data = b64encode(content.encode(charset)).decode('ascii')
INTERNALERROR> AttributeError: 'bytes' object has no attribute 'encode'
> /Users/caged/miniconda3/envs/bokeh_34/lib/python3.4/site-packages/pytest_html-1.7-py3.4.egg/pytest_html/plugin.py(74)data_uri()
-> data = b64encode(content.encode(charset)).decode('ascii')


 72     def data_uri(content, mime_type='text/plain', charset='utf-8'):
 73         if PY3:
 74  ->         data = b64encode(content.encode(charset)).decode('ascii')
 75         else:
 76             data = b64encode(content)
 77         return 'data:{0};charset={1};base64,{2}'.format(mime_type, charset, data)
````

When in pdb, just running `b64encode(content)` works great - maybe PY3 check isn't quite right? Could we check what content is instead?

davehunt:
@birdsarah could you paste more of your stack trace? I'm most interested in what content is, and where it's being provided. I assume this is one of the debug items from pytest-selenium but it's not clear which one. Note that pytest-selenium doesn't officially support Python 3.x, and it's possible it could be fixed there. That said, we probably want to support `bytes` objects in pytest-html too.

davehunt:
I was able to replicate this, and have addressed it in #28. I am still interested in where this bytes object is being created though, if you could provide more information.

davehunt:
My guess is that this is caused by pytest-selenium encoding [here](https://github.com/davehunt/pytest-selenium/blob/db3f61afa2477735a6ec8bf29737e011fd7bdc9b/pytest_selenium/pytest_selenium.py#L153) and [here](https://github.com/davehunt/pytest-selenium/blob/db3f61afa2477735a6ec8bf29737e011fd7bdc9b/pytest_selenium/pytest_selenium.py#L226). This is necessary for Python 2.x but perhaps not for 3.x? @The-Compiler do you think this should be fixed in pytest-selenium to just pass the unicode object instead of encoding it first?

The-Compiler:
[...]

PR #28, opened 2016-01-26
Added support for bytes objects in Python 3.x

davehunt:
@The-Compiler could you take a look and see if you think this is a suitable fix for #27?

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -71,7 +71,9 @@ def pytest_unconfigure(config):
 
 def data_uri(content, mime_type='text/plain', charset='utf-8'):
     if PY3:
-        data = b64encode(content.encode(charset)).decode('ascii')
+        if not isinstance(content, bytes):
+            content = content.encode(charset)
+        data = b64encode(content).decode('ascii')
     else:
         data = b64encode(content)
     return 'data:{0};charset={1};base64,{2}'.format(mime_type, charset, data)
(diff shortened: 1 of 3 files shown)


test_pytest_html.py
```diff
@@ -202,6 +202,31 @@ def pytest_runtest_makereport(item, call):
             href)
         assert link in html
 
+    @pytest.mark.xfail(
+        sys.version_info < (3, 2),
+        reason='Only affects Python versions later than 3.2',
+        run=False)
+    def test_extra_bytes(self, testdir):
+        content = bytes(str(random.random()), 'utf-8')
```
The-Compiler: This will give you a bytes-object with a random number in it - not sure how useful that is? :wink:

Also, the more idiomatic way to do this would be `str(random.random()).encode('utf-8')` - you'll get the same thing, but `.encode()`/`.decode()` are more common than the two-argument form of `bytes`/`str`

pytest_html/plugin.py
```diff
@@ -71,7 +71,9 @@ def pytest_unconfigure(config):
 
 def data_uri(content, mime_type='text/plain', charset='utf-8'):
     if PY3:
-        data = b64encode(content.encode(charset)).decode('ascii')
+        if not isinstance(content, bytes):
+            content = content.encode(charset)
```
The-Compiler: This will work, but is generally frowned upon - I'm not familiar enough with the context though to say if it'd make more sense to do it like this, or e.g. add a `data_uri_bytes`.

davehunt:
We've decided to fix this at the source. See #27.

Issue #40, opened 2016-05-06
Report Table Configuration


michael-hart:
A feature request - would it be possible to add makereport hooks that allow the user to configure columns to add, or even hide? For instance, add a column with the docstring for a test description.

michael-hart:
Possible solution to #34

davehunt:
Fixed by adding hooks in bc2e8a35554df68113cd2ae911dc396cd32cfe93 available in v1.14.0.

PR #41, opened 2016-05-06
Table Configuration

michael-hart:
I've written a basic implementation to #40. It has documentation and passes the same tests as before. However, it lacks assertions and tests to cover the new functionality.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -129,6 +129,7 @@ Then use a :code:`pytest_runtest_makereport` hook to add extras to the extra_col
       # Add ball.png as an image
       extra_col.append(
         pytest_html.extras.image('ball.png', name='Image Representation'))
+      report.extra_col = extra_col
 
 
 Screenshots
--- a/README.rst
+++ b/README.rst
@@ -87,6 +87,50 @@ conftest.py file:
               extra.append(pytest_html.extras.html('<div>Additional HTML</div>'))
           report.extra = extra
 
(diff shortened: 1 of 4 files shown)


michael-hart:
Hi, the tests are failing at present, but they can be fixed - I'd just like some confirmation that this is a desirable feature before I spend time writing further tests. Is this PR something you are interested in?

davehunt:
Sorry for the silence.. I do like the idea of columns being configurable by the user, but I haven't spent a lot of time thinking about it. This summer I'll be mentoring @RibeiroAna on several enhancements to this plugin, so I may ask her to look through your patch and help us to move it forward.

RibeiroAna:
Hi @davehunt, I'll take a look =)

davehunt:
@michael-hart are you still interested in working on this? Looking over your patch, I think a better approach would be to implement custom hooks for the table headers and columns. We could build the row as a list of cells, as you have in `_appendrow` and then call a hook such as `pytest_html_results_table_row(cells)`. We can then do the same for the header with a `pytest_html_results_table_header(cells)`. This would allow a user to implement these hooks and modify/add to the list of cells.

michael-hart:
Hi, I'm a bit busy at the moment, but I can probably take a look in a few weeks?

davehunt:
@michael-hart I decided to give it a go myself, and have a solution in #104. If you have time, please take a look and leave feedback. Closing this pull request in favour of the new approach.

PR #42, opened 2016-05-11
Undo sort by clicking a third time

rpuntaie:
I once made a pull request with an index, which you rejected. This time I changed main.js such that a third click restores the order of test execution. Please consider merging. Thanks.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -236,7 +236,7 @@ class HTMLReport(object):
         results = [html.h2('Results'), html.table([html.thead(
             html.tr([
                 html.th('Result',
-                        class_='sortable initial-sort result',
+                        class_='sortable result',
                         col='result'),
                 html.th('Test', class_='sortable', col='name'),
                 html.th('Duration',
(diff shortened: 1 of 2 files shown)


davehunt:
This removes the default of sorting by result, which brings failures to the top. I also don't think a third click to clear a sort is particularly intuitive. Didn't we discuss adding a timestamp column, which could be used to sort the tests by order of execution?

rpuntaie:
>>> This removes the default of sorting by result
This was just a preference of mine. There is no link to the third click feature. The default ordering is the one in the file. It should be possible to return to that.

davehunt:
Perhaps we could make the default sort order configurable. Note that by not specifying one, the results are likely to be in the order of execution, which is related to the order of discovery, which is unpredictable especially if tests are executed in parallel.

rpuntaie:
The user might use pytest-ordering to force a certain order of test execution. Even if the order of tests has no relevance to the tester at first, it might become important in the analysis of the test results. Anyway, it is information, which would get lost otherwise, unless one looks into the file with an editor.

davehunt:
The order in the HTML file will itself be changed in #29 so checking the source file won't help there. I don't see a convincing argument for removing the default sort order of test result, or for preserving the unsorted order of results. I would be happy to review a patch that introduces a timestamp column, or allows the default sort to be configurable via a configuration file setting.

davehunt:
I'm going to close this. Once #58 is resolved I think it could make sense to allow the default sort order to be configurable. Alternatively we could provide a hook that could be used to override the sorting of results.

Issue #37, opened 2016-03-07
Add support for automatic reruns


davehunt:
The [pytest-rerunfailures](https://pypi.python.org/pypi/pytest-rerunfailures/) plugin automatically reruns failed tests. We should include the reruns in the HTML report, including any debug information. This can help to identify if persistent failures are failing in a consistent manner. Each rerun should have its own row similar to failure but the status should be orange. The total number of reruns should be reported in the summary. Note that due to https://github.com/pytest-dev/pytest/issues/1193, if [pytest-xdist](https://pypi.python.org/pypi/pytest-xdist) is used, reruns will not be included in the report.

PR #46, opened 2016-05-25
Filter the tests results by outcome #issue38

RibeiroAna:
How it looks: when the page is first loaded, there are checked check-boxes and by checking and unchecking the boxes, the users can filter which tests they want to see. If there are no tests of one outcome, the checkbox is disabled.

image screenshot


davehunt:
@RibeiroAna you don't need to close and open new pull requests when you modify your code. If you push additional commits to the same branch, GitHub will show the changes in the open pull request. Once we're ready to merge we can squash the commits if needed.

pytest_html/plugin.py
```diff
@@ -154,8 +154,8 @@ def _appendrow(self, result, report):
             html.td(test_id, class_='col-name'),
             html.td('{0:.2f}'.format(time), class_='col-duration'),
             html.td(links_html, class_='col-links'),
-            html.td(additional_html, class_='extra')],
-            class_=result.lower() + ' results-table-row'))
+            html.td(additional_html, class_='extra', id=test_id)],
+            class_=result.lower() + ' results-table-row', id=test_id))
```
davehunt: We shouldn't have two elements with the same id.

 > The id global attribute defines a unique identifier (ID) which must be unique in the whole document.

From: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

pytest_html/plugin.py
```diff
@@ -215,21 +215,49 @@ def pytest_sessionfinish(self, session):
...
+                filter_checkbox[test_type] = html.input(type='checkbox',
+                                                        disabled='true')
+
         summary = [html.h2('Summary'), html.p(
-            '{0} tests ran in {1:.2f} seconds.'.format(
+            '{0} tests ran in {1:.2f} seconds. '.format(
                 numtests, suite_time_delta),
+            html.p('(Un)check the boxes to filter the results.'),
             html.br(),
+            filter_checkbox['Passed'],
```
davehunt: What do you think about having a list of outcomes, each item in the list could be a dictionary containing the name, class, and label. Then we could simply iterate through the list and build the summary along with the checkboxes, rather than creating a list of checkboxes ahead of time? This would remove the line break in the summary, but I think it will look okay. It would also make it easy to extend the outcomes, which might be useful for https://github.com/davehunt/pytest-html/issues/37.

davehunt:
In testing this, I'm unable to filter out errors. I suspect this is related to differences in the test id. Here are some examples for generating setup/teardown errors:

```python
@pytest.fixture
def startup_error():
    raise Exception('startup_error')


def test_startup_error(startup_error):
    pass


@pytest.fixture
def teardown_error(request):
    def fin():
        raise Exception('teardown_error')
    request.addfinalizer(fin)


def test_teardown_error(teardown_error):
    pass
```

pytest_html/plugin.py
davehunt: You could consider using [data attributes](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes) instead.

test_pytest_html.py
```diff
@@ -37,6 +48,7 @@ def assert_summary(html, tests=1, duration=None, passed=1, skipped=0, failed=0,
     assert int(re.search('(\d)+ errors', html).group(1)) == errors
     assert int(re.search('(\d)+ expected failures', html).group(1)) == xfailed
     assert int(re.search('(\d)+ unexpected passes', html).group(1)) == xpassed
+    assert_summary_checkboxes(html, passed)
```
davehunt: Rather than call another method and only check passed, it wouldn't be much additional code to check all boxes are in the HTML and have the appropriate state. You could potentially use `**kwargs` to loop over the outcomes passed to this method and reduce duplication. You might then need a dict for mapping `xfailed` to `expected failures`, etc.

davehunt: I think it may make sense to also add a test that opens the HTML report in a browser and verifies that toggling the checkboxes filters the results. This would have caught the issue I'm seeing with filtering errors.

pytest_html/plugin.py
RibeiroAna: It sounds good, what do you think about putting this list in a different file, like a "utils" class, so it could be read by the test file and by the plugin

davehunt: The test file should be able to import from the plugin module, but I'm okay with creating a separate file if you prefer it. Rather than a dictionary, I wonder if a custom object might make more sense.

pytest_html/plugin.py
```diff
@@ -215,23 +218,51 @@ def pytest_sessionfinish(self, session):
             html.title('Test Report'),
             html.style(raw(style_css)))
 
+
+        class ResultByType:
```
RibeiroAna: I have no idea if here would be the best place for a new class

pytest_html/plugin.py
```diff
@@ -149,13 +149,16 @@ def _appendrow(self, result, report):
         if report.when != 'call':
             test_id = '::'.join([report.nodeid, report.when])
 
+        logs_attribs = {"data-type-associated" : result, 
+                        "data-test-associated" : test_id};
```
davehunt: Could we use `data-test-result` and `data-test-id` instead? Also, let's make the result lowercase.

pytest_html/plugin.py
davehunt: This is fine. We could also move it to it's own file. In this case is the type the outcome? I'm a bit confused by the name of the class. This object represents each potential outcome, perhaps `Outcome` would be a better name?

pytest_html/plugin.py
```diff
@@ -215,23 +218,51 @@ def pytest_sessionfinish(self, session):
             html.title('Test Report'),
             html.style(raw(style_css)))
 
+
+        class ResultByType:
+
+            def __init__(self, label="Passed", class_html="passed",
+                         data_type_associated="Passed", amount=0):
```
davehunt: I think we should have two required arguments: outcome and total. Then label, class, and data should be the same as the outcome unless they're provided.

pytest_html/resources/main.js
```diff
@@ -190,3 +190,30 @@ function one_row_for_data() {
...
+
+    
+    for(var i = 0; i < rows.length; i++){
+        if(rows[i].getAttribute(att_name) ==
+            elem.getAttribute(att_name)) {
+            rows[i].hidden = !checkbox.checked;
+        }
+    }
+
+    for(var i = 0; i < logs.length; i++) {
```
davehunt: Could we combine these two loops?

RibeiroAna: I don't think so, that is because in the first one, we are iterating on rows.length, and in the second one in logs.length

davehunt: The inner code is identical though, so if you combined `rows` and `logs` and then iterated through them, wouldn't that work?

pytest_html/resources/main.js
```diff
@@ -190,3 +190,33 @@ function one_row_for_data() {
...
+    for(var i = 0; i < rows.length; i++){
+        if(rows[i].getAttribute(outcome_att) == outcome) {
+            rows[i].hidden = hidden;
+        }
+    }
+
+    hide_extra_html_by_attribute(hidden, outcome_att, outcome);
+}
+
+function hide_extra_html_by_attribute(hidden, attribute_key, attribute_value) {
```
RibeiroAna: @davehunt I don't think that combining those loops will help a lot, right now we have a complexity r + l where r is the size of rows and l the size of logs, in big O notation, it would be only O(r), so it is a algorithm with a linear complexity. Combining those loops, we still would have a linear  complexity (so speaking about only complexity, it would not improve the algorithm), but we would have some conditions:
=> There could be a bigger chance of accessing a index that don't exist
=> The loop would be confusing
=> The could would not be reusable
=> We would have to do many checks (such as to know which array is bigger, check if the index is not bigger than the array length, and maybe others, it smells like bug)
I split this methods in two because the second method can be used on the issue of collapsing tests logs, and also in the search issue (because we will want to collapse the logs of the methods that don't matches the search).

davehunt: We just had a quick chat about this, but let's explore grouping two rows within a `<tbody>`, and then using this for toggling display of both rows. This save us from splitting the rows using JavaScript in the first place, and also means the report looks better when JavaScript is disabled. I uploaded a rough (mostly) working example here: http://people.mozilla.org/~dhunt/tbody-report.html

pytest_html/plugin.py
```diff
@@ -141,21 +141,26 @@ def _appendrow(self, result, report):
                 log.append(html.br())
                 log.append(content)
         else:
-            log = html.div(class_='empty log')
+            log = html.div(class_='empty log', colspan="5")
```
davehunt: The colspan should be on the `<td>`, and not needed here.

pytest_html/plugin.py
```diff
@@ -141,21 +141,26 @@ def _appendrow(self, result, report):
...
             html.td(result, class_='col-result'),
             html.td(test_id, class_='col-name'),
             html.td('{0:.2f}'.format(time), class_='col-duration'),
-            html.td(links_html, class_='col-links'),
-            html.td(additional_html, class_='extra')],
-            class_=result.lower() + ' results-table-row'))
+            html.td(links_html, class_='col-links')])
+
+        rows_extra = html.tr(html.td(additional_html,
+                             class_='extra', colspan="5"))
```
davehunt: Nit: For consistency use `'` instead of `"`

pytest_html/plugin.py
```diff
@@ -215,23 +220,55 @@ def pytest_sessionfinish(self, session):
...
+                self.label = label or outcome
+                self.class_html = class_html or outcome
+                self.total = total
+                self.test_result = test_result or outcome
+
+                self.generate_checkbox()
+                self.generate_summary_item()
+
+            def generate_checkbox(self):
+                data_test_result = {"data-test-result":
```
davehunt: Nit: Use `'` instead of `"` for consistency.

pytest_html/plugin.py
```diff
@@ -215,23 +220,55 @@ def pytest_sessionfinish(self, session):
...
+                                    self.test_result.lower()}
+                if(self.total > 0):
```
davehunt: There's no need for the brackets in Python logic blocks.

pytest_html/plugin.py
```diff
@@ -215,23 +220,55 @@ def pytest_sessionfinish(self, session):
...
+                    self.checkbox = html.input(type='checkbox',
+                                               checked='true',
+                                               onChange='filter_table(this)',
+                                               name='filter_checkbox',
+                                               **data_test_result)
+                else:
+                    self.checkbox = html.input(type='checkbox',
+                                               disabled='true')
```
davehunt: Intead of having two different values for `checkbox` you could have one, but make the value of `disabled` dependent on the number of tests with the outcome. For example:

```python
self.checkbox = html.input(
    type='checkbox',
    disabled='true' if self.total > 0 else 'false',
    ...)
```

The remaining attributes will be present for the disabled checkbox, but that won't matter because it's disabled.

pytest_html/plugin.py
```diff
@@ -215,23 +220,55 @@ def pytest_sessionfinish(self, session):
...
+
+            def generate_summary_item(self):
+                self.summary_item = html.span('{0} {1}'.
+                                              format(self.total, self.label),
+                                              class_=self.class_html)
+
+        outcomes = [Outcome('passed', self.passed),
+                    Outcome('skipped', self.skipped),
+                    Outcome('failed', self.failed),
+                    Outcome('error', self.errors, 'errors'),
```
davehunt: I think this will be clearer if we include the keyword for the argument: `label='errors'`

pytest_html/plugin.py
```diff
@@ -215,23 +220,55 @@ def pytest_sessionfinish(self, session):
...
+                    Outcome('xfailed', self.xfailed,
+                            'expected failures', 'xfailed', 'skipped'),
```
davehunt: The `test_result` argument is not needed as it matches the outcome. You can avoid specifying it by passing the keywords for the optional arguments: `Outcome('xfailed', self.xfailed, label='expected failures', class_html='skipped'`

test_pytest_html.py
```diff
@@ -25,6 +25,23 @@ def run(testdir, path='report.html', *args):
     return result, html
 
 
+def assert_summary_filter(html, test_outcome='Passed', test_outcome_number=1):
```
davehunt: This is always called with an outcome and total, so I don't think they need to be optional arguments.

test_pytest_html.py
```diff
@@ -25,6 +25,23 @@ def run(testdir, path='report.html', *args):
...
+    cb_active = ("<input checked=\"true\""
+                 " data-test-result=\"{0}\""
+                 .format(test_outcome))
+    cb_unactive = ("<input disabled=\"true\"")
```
davehunt: If you had identical checkbox HTML other than the value of disabled, you could do a regex match with a group for the disabled value. This would mean you'd just have one regular expression, but would be able to check the state of disabled.

test_pytest_html.py
```diff
@@ -25,6 +25,23 @@ def run(testdir, path='report.html', *args):
...
+    table_test_result_outcome = ("tbody class=\"{0} "
+                                 .format(test_outcome))
+    if(test_outcome_number != 0):
```
davehunt: No need for the brackets.

test_pytest_html.py
```diff
@@ -25,6 +25,23 @@ def run(testdir, path='report.html', *args):
...
+        assert str(re.search(cb_active, html).group()) == cb_active
+        assert len(
+                    re.findall(
+                               table_test_result_outcome,
+                               html)) == test_outcome_number
```
davehunt: This is a worthwhile check, but it's not really related to the filter. Perhaps move it into `assert_summary`.

davehunt:
The sort is only working for me every other click. Also, could you add a message to the page when all results have been filtered out? I've added a few other comments, but this is looking great!

pytest_html/plugin.py
RibeiroAna: @davehunt For some reason, whenever the attribute disabled is in the checkbox, doesn't matter its value (either true or false, disabled or '', whatever), the checkbox is disabled. Not just me faces this issue: http://stackoverflow.com/questions/3557266/cant-set-disabled-false-javascript), so seem we can't use that like that

davehunt: Ah, right, it's a boolean attribute that doesn't require a value. You could probably do it by making `data_test_result` more generic, such as:

```python
checkbox_kwargs = {
    'data-test-result': self.test_result.lower()}

if self.total > 0:
    checkbox_kwargs['disabled'] = 'true'

self.checkbox = html.input(
    type='checkbox',
    checked='true',
    onChange='filter_table(this)',
    name='filter_checkbox',
    **checkbox_kwargs)
```

test_pytest_html.py
```diff
@@ -25,22 +25,47 @@ def run(testdir, path='report.html', *args):
...
-    assert int(re.search('(\d)+ errors', html).group(1)) == errors
-    assert int(re.search('(\d)+ expected failures', html).group(1)) == xfailed
-    assert int(re.search('(\d)+ unexpected passes', html).group(1)) == xpassed
+        assert float(tests_duration.group(1)) >= float(duration)
 
+    # Asserts number of rows of the table
+    assert len(re.findall('tbody', html)) or 0 == tests
 
-class TestHTML:
+    # Asserts by outcome
```
RibeiroAna: Hi @davehunt, new commit... I repeat less the code, writing the whole checking code just once... I Just have a question, is there anyway to not repeat the parameter here?

test_pytest_html.py
```diff
@@ -25,22 +25,48 @@ def run(testdir, path='report.html', *args):
...
+    assert_results_by_outcome(html, 'passed', passed)
```
RibeiroAna: @davehunt by doing that I don't have to write all the individual checks again

test_pytest_html.py
```diff
@@ -25,22 +25,48 @@ def run(testdir, path='report.html', *args):
...
+def assert_results(html, tests=1, duration=None, passed=1, skipped=0, failed=0,
                    errors=0, xfailed=0, xpassed=0):
-    m = re.search('(\d)+ tests ran in ([\d,.])+ seconds', html)
-    assert int(m.group(1)) == tests
+    # Asserts total amount of tests
+    total_tests = re.search('(\d)+ tests ran', html)
+    assert int(total_tests.group(1)) == tests
+
+    # Asserts tests running duration
+    tests_duration = re.search('([\d,.])+ seconds', html)
```
davehunt: We could move this into the if block, so we don't do the search unless we're verifying the value.

test_pytest_html.py
```diff
@@ -25,22 +25,48 @@ def run(testdir, path='report.html', *args):
...
-    assert int(re.search('(\d)+ passed', html).group(1)) == passed
-    assert int(re.search('(\d)+ skipped', html).group(1)) == skipped
-    assert int(re.search('(\d)+ failed', html).group(1)) == failed
-    assert int(re.search('(\d)+ errors', html).group(1)) == errors
-    assert int(re.search('(\d)+ expected failures', html).group(1)) == xfailed
-    assert int(re.search('(\d)+ unexpected passes', html).group(1)) == xpassed
+        assert float(tests_duration.group(1)) >= float(duration)
 
+    # Asserts number of rows of the table
+    assert len(re.findall('tbody', html)) or 0 == tests
```
davehunt: The `or 0` should not be needed, as `findall` will return an empty list if there are no matches.

test_pytest_html.py
```diff
@@ -25,22 +25,48 @@ def run(testdir, path='report.html', *args):
     return result, html
 
 
-def assert_summary(html, tests=1, duration=None, passed=1, skipped=0, failed=0,
+def assert_results_by_outcome(html, test_outcome, test_outcome_number,
+                              label=None):
+    # Asserts if the test number of this outcome in the summary matches
+    regex_summary = '(\d)+ {0}'.format(label or test_outcome)
+    int(re.search(regex_summary, html).group(1)) == test_outcome_number
```
davehunt: You're missing an assert here.

test_pytest_html.py
```diff
@@ -25,22 +25,48 @@ def run(testdir, path='report.html', *args):
...
+
+    # Asserts if the generated checkboxes of this outcome matches
+    regex_checkbox = ('<input checked="true" data-test-result="{0}"'
+                      .format(test_outcome))
+    if test_outcome_number == 0:
+        regex_checkbox += ' disabled="true"'
+    assert str(re.search(regex_checkbox, html).group()) == regex_checkbox
```
davehunt: No need to do a string compare here, you can just assert that `re.search ... is not None`

test_pytest_html.py
```diff
@@ -25,22 +25,48 @@ def run(testdir, path='report.html', *args):
...
+
+    # Asserts if the generated checkboxes of this outcome matches
```
davehunt: Need to update this comment.

test_pytest_html.py
```diff
@@ -25,22 +25,48 @@ def run(testdir, path='report.html', *args):
...
+    regex_table = ('tbody class=\"{0} '.format(test_outcome))
+    assert len(re.findall(regex_table, html)) or 0 == test_outcome_number
```
davehunt: No need for the `or 0`

pytest_html/resources/main.js
```diff
@@ -110,21 +109,21 @@ function sort(items, key_func, reversed) {
...
-        return elem.childNodes[col_index].firstChild.data.toLowerCase();
+        return elem.childNodes[1].childNodes[col_index].firstChild.data.toLowerCase();
     };
 }
 
 function key_num(col_index) {
     return function(elem) {
-        return parseFloat(elem.childNodes[col_index].firstChild.data);
+        return parseFloat(elem.childNodes[1].childNodes[col_index].firstChild.data);
     };
```
RibeiroAna: @davehunt the problem was that before we had just one child (like, it was <td><tr></tr></td>, but now it is <tbody><td>first one about the row</td><td>logs</td></tr>), so there was one more level to access.

davehunt: Aha! I knew you'd work it out. 😀

test_pytest_html.py
```diff
@@ -25,22 +25,45 @@ def run(testdir, path='report.html', *args):
     return result, html
 
 
-def assert_summary(html, tests=1, duration=None, passed=1, skipped=0, failed=0,
+def assert_results_by_outcome(html, test_outcome, test_outcome_number,
+                              label=None):
+    # Asserts if the test number of this outcome in the summary is correct
+    regex_summary = '(\d)+ {0}'.format(label or test_outcome)
+    int(re.search(regex_summary, html).group(1)) == test_outcome_number
```
davehunt: Still missing an `assert` here.

test_pytest_html.py
```diff
@@ -25,22 +25,45 @@ def run(testdir, path='report.html', *args):
...
+
+    # Asserts if the generated checkbox of this outcome is correct
+    regex_checkbox = ('<input checked="true" data-test-result="{0}"'
+                      .format(test_outcome))
+    if test_outcome_number == 0:
+        regex_checkbox += ' disabled="true"'
+    assert str(re.search(regex_checkbox, html).group()) != None
```
davehunt: No need to return a string here, and as the flake8 checks point out, comparison to `None` should be `is not None`.

davehunt:
This is looking great, @RibeiroAna! Could you add a message to the table when the user filters all the results out? Just something like "No results? Try removing some of the filters." In a later enhancement we could even add a "clear all" link to the message to remove any applied filters.

RibeiroAna:
Hi @davehunt , I just put the "not found" message. We could open a new issue to "clear all" function and merge that one =)

davehunt:
> We could open a new issue to "clear all" function and merge that one =)

I agree. I'll look at your latest changes first thing tomorrow. Thanks!

pytest_html/resources/main.js
```diff
@@ -160,33 +156,31 @@ function toggle_sort_states(elem) {
...
-                new_row.appendChild(td_elem);
-                td_elem.colSpan=5;
-            } else {
-                td_elem.parentNode.removeChild(td_elem);
-            }
-        });
-    });
+function is_all_rows_hidden() {
+    var rows = find_all('.results-table-row');
+    if(rows.length == 0)
```
davehunt: Could you use a [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) here, to reduce the array to only rows that are not hidden. Then you can just return the result of checking the length against zero.

davehunt:
@RibeiroAna just three comments to address, and then I think we're ready to merge. Rather than waiting on me, please feel free to move onto #49 (which I think is a great idea), or #37.

PR #54, opened 2016-06-28
Tag 1.9.0

RibeiroAna:


davehunt:
I've released v1.9.0 here: https://pypi.python.org/pypi/pytest-html/1.9.0

PR #57, opened 2016-07-03
Split pytest_sessionfinish into generate and save methods.

karandesai-96:
This PR exercises a minor refactor in `HTMLReport` class. The performance of plugin remains unaltered. 

#### Then:
* `pytest_sessionfinish` method used to open the logfile, write all the html content, and finally close and save it.

#### Now:
* This method has been split to two separate methods: `_generate_report` and `_save_report`.
* `_generate_report` returns a unicode string with all the content.
* `_save_report` accepts this as a param and writes it to the logfile.
* `pytest_sessionfinish` calls both of these methods one after another.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -312,7 +312,7 @@ class HTMLReport(object):
 
         doc = html.html(head, body)
 
-        unicode_doc = doc.unicode(indent=2)
+        unicode_doc = u'<!DOCTYPE html>\n{0}'.format(doc.unicode(indent=2))
         if PY3:
             # Fix encoding issues, e.g. with surrogates
             unicode_doc = unicode_doc.encode('utf-8',
@@ -325,7 +325,6 @@ class HTMLReport(object):
             os.makedirs(os.path.dirname(self.logfile))
         logfile = open(self.logfile, 'w', encoding='utf-8')
 
-        logfile.write('<!DOCTYPE html>')
         logfile.write(report_content)
         logfile.close()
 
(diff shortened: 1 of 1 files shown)


karandesai-96:
Hi @davehunt I am using `pytest-html` as a library and not a plugin, because we have a slightly different use case. We do not want the report saved in a local html file. We are directly uploading raw html to the organization wiki. Hence, `_generate_report` remains same, but we subclass `HTMLReport` and override its `_save_report` method accordingly.

There might be a little customization which an individual would be prefer and in my opinion, this refactor will allow a little more flexibility during sub classing.

pytest_html/plugin.py
```diff
@@ -328,16 +312,40 @@ def generate_summary_item(self):
...
             unicode_doc = unicode_doc.decode('utf-8')
-        logfile.write(unicode_doc)
+        return unicode_doc
+
+    def _save_report(self, report_content):
+        if not os.path.exists(os.path.dirname(self.logfile)):
+            os.makedirs(os.path.dirname(self.logfile))
+        logfile = open(self.logfile, 'w', encoding='utf-8')
+
+        logfile.write('<!DOCTYPE html>')
```
davehunt: This means that `_generate_report` returns a report that doesn't include this doctype declaration. I think it makes sense to include this in `_generate_report`. Otherwise, I'm happy with this change.

karandesai-96: @davehunt I have modified this change in subsequent commit. Thanks for the review !

davehunt: The tests are now failing, could you take a look?

karandesai-96: I rectified a little typo. The tests are passing now.

Issue #58, opened 2016-07-04
Order table without JS


RibeiroAna:
This is one enhancement related to issue #29

davehunt:
To add a little more detail, this issue relates to sorting the results in the HTML content itself to avoid an initial JavaScript sort when the page loads. This improves performance when there are hundreds or thousands of results, and also means that the results are displayed in a useful order when JavaScript is not permitted.

PR #60, opened 2016-07-05
Optimize results sorting

### LINKED ISSUE #59: Sorting is working extremely slow in Firefox
Firefox 47.0 Ubuntu 16.04 LTS

Browser is asking to stop script when trying to open big html report.

Here is the sample report with 2558 tests: https://r-ci.tk/stuff/pytest-old.html

redixin:
Clean up table before calling appendElement for each element.

Actually table is completely removed and created a new one. This is the
fastest way to have clean table.

This approach is ~10% faster in chromium and ~400% faster in Firefox.

Here is sample report pages:

With this patch: https://r-ci.tk/stuff/pytest.html
Without this patch: https://r-ci.tk/stuff/pytest-old.html

Closes #59

### DIFF
--- a/pytest_html/resources/main.js
+++ b/pytest_html/resources/main.js
@@ -82,11 +82,18 @@ function sort_table(clicked, key_func) {
     var rows = find_all('.results-table-row');
     var reversed = !clicked.classList.contains('asc');
     var sorted_rows = sort(rows, key_func, reversed);
-
-    var parent = document.getElementById('results-table');
+    /* Whole table is removed here because browsers acts much slower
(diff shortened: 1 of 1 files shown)


andreykurilin:
Big +1. Results of http://www.webpagetest.org for FireFox:
without this patch: First View (21.871s)
after this patch: First View (9.842s)

but it's still too slow :(

davehunt:
Thanks for the patch! With #58 resolved we'll avoid the initial sort via JavaScript too, which should help with the performance.

PR #61, opened 2016-07-05
Order table without JS

RibeiroAna:
related to issue #58

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -12,6 +12,7 @@ import pkg_resources
 import platform
 import sys
 import time
+import bisect
 
 import pytest
 from py.xml import html, raw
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,11 @@
 Release Notes
 -------------
 
+**1.9.0 (2016-06-28)**
+
+* Added support to rerun plugins
+* Added a feature to filter tests by its outcome
+
```
davehunt: Can you rebase from upstream/master so we don't include this change?

pytest_html/plugin.py
```diff
@@ -83,97 +84,117 @@ def __init__(self, logfile):
...
+            self.row_table = html.tr([
+                html.td(self.outcome, class_='col-result'),
+                html.td(self.test_id, class_='col-name'),
+                html.td('{0:.2f}'.format(self.time), class_='col-duration'),
+                html.td(self.links_html, class_='col-links')])
+
+            self.row_extra = html.tr(html.td(self.additional_html,
+                                     class_='extra', colspan='5'))
+
+        def __cmp__(self, other):
```
davehunt: I suspect this won't give us the desired sort order. Sorting by outcome currently uses the following order: Error, Failed, XFailed, XPassed, Skipped, Passed as [defined here](https://github.com/pytest-dev/pytest-html/blob/master/pytest_html/resources/main.js#L131-L132) (we should also include rerun in there). I'm also not sure about the sorting by name within outcomes - it looks like this is skipped if the outcomes are already sorted appropriately? As far as I can tell, it's also not something that's currently possible in the report using JavaScript so we wouldn't be able to restore this sort order without refreshing the page.

pytest_html/plugin.py
```diff
@@ -83,97 +84,117 @@ def __init__(self, logfile):
...
-                    line = line.decode('utf-8')
-                separator = line.startswith('_ ' * 10)
-                if separator:
-                    log.append(line[:80])
-                else:
-                    exception = line.startswith("E   ")
-                    if exception:
-                        log.append(html.span(raw(escape(line)),
-                                             class_='error'))
+    class TestReport:
```
davehunt: I'd prefer if we named this `TestResult` as it represents a single result rather than a full report.

pytest_html/plugin.py
```diff
@@ -83,97 +84,117 @@ def __init__(self, logfile):
...
-            html.td(links_html, class_='col-links')])
-
-        rows_extra = html.tr(html.td(additional_html,
-                             class_='extra', colspan='5'))
-
-        self.test_logs.append(html.tbody(rows_table, rows_extra,
-                                         class_=result_name.lower() +
-                                         ' results-table-row'))
+    def _appendrow(self, outcome, report):
+        test = self.TestReport(outcome, report)
```
davehunt: Call this `result` instead of `test`.

pytest_html/plugin.py
```diff
@@ -83,97 +84,117 @@ def __init__(self, logfile):
         logfile = os.path.expanduser(os.path.expandvars(logfile))
         self.logfile = os.path.abspath(logfile)
         self.test_logs = []
+        self.tests = []
```
davehunt: Call this `self.results`.

pytest_html/plugin.py
```diff
@@ -83,97 +84,117 @@ def __init__(self, logfile):
...
+                     'XPassed': 4, 'Skipped': 5, 'Passed': 6}
+            if order[self.outcome] > order[other.outcome]:
+                return False
+            elif order[self.outcome] < order[other.outcome]:
+                return True
+            else:
+                if self.test_id > other.test_id:
+                    return False
+                elif self.test_id <= other.test_id:
+                    return True
```
davehunt: I don't think we should have a secondary sort here, even though it makes a lot of sense to do so. The problem is that this is not an equivalent of anything that can currently be achieved in the report. If you currently sort by outcome, the tests within each outcome actually varies. This also means there's no way to restore the original sort order except by refreshing the page. I think we'd need to have support for sorting by multiple columns in order to introduce this as a default order. This simplifies your code to an if statement and a return statement - no need for any elif/else.

pytest_html/plugin.py
```diff
@@ -83,97 +84,117 @@ def __init__(self, logfile):
...
+                html.td(self.outcome, class_='col-result'),
+                html.td(self.test_id, class_='col-name'),
+                html.td('{0:.2f}'.format(self.time), class_='col-duration'),
+                html.td(self.links_html, class_='col-links')])
+
+            self.row_extra = html.tr(html.td(self.additional_html,
+                                     class_='extra', colspan='5'))
+
+        def __lt__(self, other):
+            order = {'Error': 0, 'Failed': 1, 'Rerun': 2, 'XFailed': 3,
```
davehunt: This doesn't need to be a dictionary. It can be a tuple, which you can then use `order.index(self.outcome)`.

pytest_html/plugin.py
```diff
@@ -83,97 +84,112 @@ def __init__(self, logfile):
...
+            self.row_extra = html.tr(html.td(self.additional_html,
+                                     class_='extra', colspan='5'))
+
+        def __lt__(self, other):
+            order = ('Error', 'Failed', 'Rerun', 'XFailed',
+                     'XPassed', 'Skipped', 'Passed')
+            if order.index(self.outcome) >= order.index(other.outcome):
+                return False
+            elif order.index(self.outcome) < order.index(other.outcome):
+                return True
```
davehunt: This can be simplified further as:

```python
def __lt__(self, other):
    order = ('Error', 'Failed', 'Rerun', 'XFailed', 'XPassed', 'Skipped', 'Passed')
    return order.index(self.outcome) < order.index(other.outcome)
```

davehunt:
@RibeiroAna is this ready for a final review?

RibeiroAna:
Yes @davehunt

PR #63, opened 2016-07-11
Added code to create an apart file to CSS

### LINKED ISSUE #62: Create a separate CSS file by default for improved CSP support
Because of CSP restrictions on inline CSS, the CSS file should be written apart

RibeiroAna:
Closes #62

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ Then run your tests with:
 
 .. code-block:: bash
 
-$ py.test --html=report.html
+   $ py.test --html=report.html
 
 
 Creating a a self-contained report
@@ -58,7 +58,7 @@ Creating a a self-contained report
 
 In order to respect the `Content Security Policy (CSP)
 <https://developer.mozilla.org/docs/Web/Security/CSP>`_,
-several assets such as CSS and images' are stored separately by default.
+several assets such as CSS and images are stored separately by default.
 You can alternatively create a self-contained report, which can be more
 convenient when sharing your results. This can be done in the following way:
 
--- a/README.rst
+++ b/README.rst
@@ -50,20 +50,22 @@ Then run your tests with:
 
 .. code-block:: bash
 
-  $ py.test --html=report.html
+$ py.test --html=report.html
 
-  Creating a single file to the project
-  -------------------------------------
 
-  In order to respect the `Content Security Policy (CSP)
-   <https://developer.mozilla.org/docs/Web/Security/CSP>`_,
-  several assets such as CSS and images' are stored separately by default.
-  You can alternatively create a self-contained report, which can be more
-  convenient when sharing your results. This can be done in the following way:
+Creating a a self-contained report
+----------------------------------
 
-  .. code-block:: bash
+In order to respect the `Content Security Policy (CSP)
+<https://developer.mozilla.org/docs/Web/Security/CSP>`_,
+several assets such as CSS and images' are stored separately by default.
+You can alternatively create a self-contained report, which can be more
+convenient when sharing your results. This can be done in the following way:
 
-      $ py.test --html=report.html --self-contained-html
+.. code-block:: bash
+
+   $ py.test --html=report.html --self-contained-html
+
 
 Enhancing reports
 -----------------
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ Then run your tests with:
 
 .. code-block:: bash
 
-   $ py.test --html=report.html
+  $ py.test --html=report.html
 
 
 Creating a a self-contained report
--- a/README.rst
+++ b/README.rst
@@ -52,8 +52,7 @@ Then run your tests with:
 
   $ py.test --html=report.html
 
-
-Creating a a self-contained report
+Creating a self-contained report
 ----------------------------------
 
 In order to respect the `Content Security Policy (CSP)
@@ -65,7 +64,6 @@ convenient when sharing your results. This can be done in the following way:
 .. code-block:: bash
 
    $ py.test --html=report.html --self-contained-html
-
 
 Enhancing reports
 -----------------
@@ -101,7 +99,6 @@ conftest.py file:
               # only add additional html on failure
               extra.append(pytest_html.extras.html('<div>Additional HTML</div>'))
           report.extra = extra
-
 
 Screenshots
 -----------
--- a/README.rst
+++ b/README.rst
@@ -52,6 +52,19 @@ Then run your tests with:
 
   $ py.test --html=report.html
 
+  Creating a single file to the project
+  -------------------------------------
+
+  In order to respect the `Content Security Policy (CSP)
+   <https://developer.mozilla.org/docs/Web/Security/CSP>`_,
+  several assets such as CSS and images' are stored separately by default.
+  You can alternatively create a self-contained report, which can be more
+  convenient when sharing your results. This can be done in the following way:
+
+  .. code-block:: bash
+
+      $ py.test --html=report.html --self-contained-html
+
 Enhancing reports
 -----------------
 
@@ -86,17 +99,6 @@ conftest.py file:
               # only add additional html on failure
               extra.append(pytest_html.extras.html('<div>Additional HTML</div>'))
           report.extra = extra
-
-Creating a single file to the project
--------------------------------------
-
-In order to respect the `Content Security Policy (CSP) <https://developer.mozilla.org/docs/Web/Security/CSP>`_, the CSS file is created apart by default. But if it is necessary to you create a single file to the whole report, you may run the file by the following way:
-
-.. code-block:: bash
-
-    $ py.test --html=report.html --self-contained-html
-
-So, the CSS will be inline and the whole report will be created on a single file.
 
 
 Screenshots
--- a/README.rst
+++ b/README.rst
@@ -87,6 +87,10 @@ conftest.py file:
               extra.append(pytest_html.extras.html('<div>Additional HTML</div>'))
           report.extra = extra
 
+Single files
+------------
+In order to protect the `Content Security Policy (CSP)<https://developer.mozilla.org/docs/Web/Security/CSP>`, the CSS file is created apart by default. But if it is necessary to you create a single file to the whole report, you may use the option `--self-contained-html` when running the project, so the CSS will be inline.
+
 Screenshots
 -----------
 
--- a/README.rst
+++ b/README.rst
@@ -87,9 +87,17 @@ conftest.py file:
               extra.append(pytest_html.extras.html('<div>Additional HTML</div>'))
           report.extra = extra
 
-Single files
-------------
-In order to protect the `Content Security Policy (CSP)<https://developer.mozilla.org/docs/Web/Security/CSP>`, the CSS file is created apart by default. But if it is necessary to you create a single file to the whole report, you may use the option `--self-contained-html` when running the project, so the CSS will be inline.
+Creating a single file to the project
+-------------------------------------
+
+In order to respect the `Content Security Policy (CSP) <https://developer.mozilla.org/docs/Web/Security/CSP>`_, the CSS file is created apart by default. But if it is necessary to you create a single file to the whole report, you may run the file by the following way:
+
+.. code-block:: bash
+
+    $ py.test --html=report.html --self-contained-html
+
+So, the CSS will be inline and the whole report will be created on a single file.
+
 
 Screenshots
 -----------
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -45,7 +45,7 @@ def pytest_addoption(parser):
                     help='create a self-contained html file containing all ' +
                     'necessary styles, scripts, and images - this means ' +
                     'that the report may not render or function where CSP ' +
-                    'restrictions are in place (see' +
+                    'restrictions are in place (see ' +
                     'https://developer.mozilla.org/docs/Web/Security/CSP)')
 
 
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -41,6 +41,9 @@ def pytest_addoption(parser):
     group.addoption('--html', action='store', dest='htmlpath',
                     metavar='path', default=None,
                     help='create html report file at given path.')
+    group.addoption('--self-contained-html', action="store", dest='self_contained_html', default=False,
+        help='If it is true, asserts that the HTML will not ' +
+             'depend of external files, risk of breaking CSP. DEFAULT: False')
 
 
 def pytest_configure(config):
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -322,10 +322,13 @@ class HTMLReport(object):
         return unicode_doc
 
     def _save_report(self, report_content):
-        if not os.path.exists(os.path.dirname(self.logfile)):
-            os.makedirs(os.path.dirname(self.logfile))
+        dir_name = os.path.dirname(self.logfile)
+        if not os.path.exists(dir_name):
+            os.makedirs(dir_name)
         logfile = open(self.logfile, 'w', encoding='utf-8')
-        style_file = open('style.css', 'w', encoding='utf-8')
+        style_path = dir_name + '/style.css'
+
+        style_file = open(style_path, 'w', encoding='utf-8')
 
         logfile.write(report_content)
         style_file.write(self.style_css)
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -41,9 +41,11 @@ def pytest_addoption(parser):
     group.addoption('--html', action='store', dest='htmlpath',
                     metavar='path', default=None,
                     help='create html report file at given path.')
-    group.addoption('--self-contained-html', action="store", dest='self_contained_html', default=False,
-        help='If it is true, asserts that the HTML will not ' +
-             'depend of external files, risk of breaking CSP. DEFAULT: False')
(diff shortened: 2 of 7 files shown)


pytest_html/plugin.py
```diff
@@ -211,15 +211,16 @@ def _generate_report(self, session):
...
         if PY3:
-            style_css = style_css.decode('utf-8')
+            self.style_css = self.style_css.decode('utf-8')
 
         head = html.head(
             html.meta(charset='utf-8'),
             html.title('Test Report'),
-            html.style(raw(style_css)))
+            html.link(href='style.css', rel='stylesheet', type='text/css'),
+            html.style(raw(self.style_css)))
```
davehunt: I'm not sure if we should still include the inline CSS if we're linking to an external file.. this also makes it hard to test.

pytest_html/plugin.py
```diff
@@ -324,9 +325,13 @@ def _save_report(self, report_content):
         if not os.path.exists(os.path.dirname(self.logfile)):
             os.makedirs(os.path.dirname(self.logfile))
         logfile = open(self.logfile, 'w', encoding='utf-8')
+        style_file = open('style.css', 'w', encoding='utf-8')
```
davehunt: This puts the `style.css` in the working directory, which means if the HTML report is stored anywhere else the CSS file will not be found. We should be sure to always place it alongside the HTML file.

pytest_html/plugin.py
RibeiroAna: That was what I was discussing first, because to some users maybe can be useful to have the a single file, so I think we could have an option or something to decide if the user wants to have all the report in a single file, or in multiple files (we could implement that to the JS as well).

davehunt:
I tested this by adding a meta tag to the HTML file that reflects the [defaults applied by Jenkins](https://wiki.jenkins-ci.org/display/JENKINS/Configuring+Content+Security+Policy#ConfiguringContentSecurityPolicy-TheDefaultRuleSet) and removing the inline CSS:

```patch
diff --git a/pytest_html/plugin.py b/pytest_html/plugin.py
index a91d03e..125b301 100644
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -218,9 +218,11 @@ class HTMLReport(object):
 
         head = html.head(
             html.meta(charset='utf-8'),
+            html.meta(**{
+                'http-equiv': 'Content-Security-Policy',
+                'content': "sandbox; default-src 'none'; img-src 'self'; style-src 'self'"}),
             html.title('Test Report'),
-            html.link(href='style.css', rel='stylesheet', type='text/css'),
-            html.style(raw(self.style_css)))
+            html.link(href='style.css', rel='stylesheet', type='text/css'))
 
         class Outcome:
```

It works well, so we just need to address the few comments and also decide how to let the users decide when they want a report with CSS inline. I'm open to suggestions for this... one idea is to add a command line option such as `--use-inline-css`, where we change the default to rely on the external file (our justification can be to support CSP by default). A potentially more user friendly option could be something like `--self-contained-html`, which would set all necessary options to ensure the HTML file did not depend on external files at the risk of breaking CSP rules. We could indeed add both, with the latter being an alias for the former and any additional options we may add to be CSP compliant.

pytest_html/plugin.py
davehunt: We could do the same with JavaScript, but by default Jenkins disallows all JavaScript regardless of it being inline or in an external file. Consider splitting the JavaScript into it's own file as outside of the scope of #29, but feel free to work on it once the main work is done.

RibeiroAna: I was thinking to let it duplicated for now and keep working on the issue #29, and then create a new issue to create either a single HTML file, or many files (the CSS and JS apart) and then remove this duplication.

davehunt: I think we should deal with it here, it just requires us to come up with a reasonable plan. I've made some suggestions here: https://github.com/pytest-dev/pytest-html/pull/63#issuecomment-231792568

pytest_html/plugin.py
```diff
@@ -41,6 +41,9 @@ def pytest_addoption(parser):
     group.addoption('--html', action='store', dest='htmlpath',
                     metavar='path', default=None,
                     help='create html report file at given path.')
+    group.addoption('--self-contained-html', action="store", dest='self_contained_html', default=False,
```
RibeiroAna: @davehunt okay, I think this adds an option, but how can I access this option latter in the code

davehunt: You can access it from `config` such as `config.getoption('self_contained_html')`. The `config` is available through many objects such as `request`, `session`, etc.

test_pytest_html.py
```diff
@@ -44,7 +44,17 @@ def assert_results_by_outcome(html, test_outcome, test_outcome_number,
...
+                   self_contained_html=False):
+
+    # Assert CSS link or inline parameter
+    if self_contained_html:
+        regex_css_inline = '<style>'
+        assert re.search(regex_css_inline, html) is not None
+    else:
+        regex_css_link = '<link href='
+        assert re.search(regex_css_link, html) is not None
+
```
RibeiroAna: @davehunt so, my idea was creating this in assert_results, but now I think the better idea is updating only the test resources

pytest_html/plugin.py
```diff
@@ -41,9 +41,11 @@ def pytest_addoption(parser):
     group.addoption('--html', action='store', dest='htmlpath',
                     metavar='path', default=None,
                     help='create html report file at given path.')
-    group.addoption('--self-contained-html', action="store", dest='self_contained_html', default=False,
-        help='If it is true, asserts that the HTML will not ' +
-             'depend of external files, risk of breaking CSP. DEFAULT: False')
+    group.addoption('--self-contained-html', action="store",
```
davehunt: I think `action` should be `store_true` so the value is `True` when this option is provided on the command line. Also, please use single quotes wherever possible.

pytest_html/plugin.py
```diff
@@ -41,9 +41,11 @@ def pytest_addoption(parser):
...
+                    dest='self_contained_html', default=False,
```
davehunt: `dest` is not needed as this will be the default name based on the argument name. If you use `store_true` as mentioned above then the default will be `False` so it's also not needed.

pytest_html/plugin.py
```diff
@@ -41,9 +41,11 @@ def pytest_addoption(parser):
...
+                    help='If it is true, asserts that the HTML will not ' +
+                    'depend of external files, risk of breaking CSP. ' +
+                    'DEFAULT: False')
```
davehunt: I think the help text should be something like:

> create a self-contained html file containing all necessary styles, scripts, and images - this means that the report may not render or function where CSP restrictions are in place (see https://developer.mozilla.org/en-US/docs/Web/Security/CSP)

pytest_html/plugin.py
```diff
@@ -211,15 +216,20 @@ def _generate_report(self, session):
...
+        self.style_css = pkg_resources.resource_string(
             __name__, os.path.join('resources', 'style.css'))
         if PY3:
-            style_css = style_css.decode('utf-8')
+            self.style_css = self.style_css.decode('utf-8')
+
+        html_css = html.link(href='style.css', rel='stylesheet',
+                             type='text/css')
+        if session.config.getoption('self_contained_html'):
+            html_css = html.style(raw(self.style_css))
```
davehunt: 👍

pytest_html/plugin.py
```diff
@@ -320,14 +330,20 @@ def generate_summary_item(self):
...
+        dir_name = os.path.dirname(self.logfile)
+        if not os.path.exists(dir_name):
+            os.makedirs(dir_name)
         logfile = open(self.logfile, 'w', encoding='utf-8')
-
         logfile.write(report_content)
         logfile.close()
 
+        if not self_contained_html:
+            style_path = dir_name + '/style.css'
```
davehunt: This won't work on Windows where the path separator is `\`. Use [`os.path.join`](https://docs.python.org/2/library/os.path.html#os.path.join)

pytest_html/plugin.py
```diff
@@ -320,14 +330,20 @@ def generate_summary_item(self):
...
+            style_file = open(style_path, 'w', encoding='utf-8')
+            style_file.write(self.style_css)
+            style_file.close()
```
davehunt: Could you use the `with` keyword so that we close the file even in the case of an exception? See the last example at https://docs.python.org/2/tutorial/inputoutput.html#methods-of-file-objects

test_pytest_html.py
```diff
@@ -174,18 +174,34 @@ def test_create_report_path(self, testdir):
         assert result.ret == 0
         assert_results(html)
 
+    def test_resources_inline_css(self, testdir):
+        testdir.makepyfile('def test_pass(): pass')
+        result, html = run(testdir, 'report.html',
+                           '--self-contained-html', 'True')
```
davehunt: With the changes mentioned already, the `True` can be dropped.

test_pytest_html.py
```diff
@@ -174,18 +174,34 @@ def test_create_report_path(self, testdir):
...
+        assert result.ret == 0
+        for resource in ['style.css', 'main.js']:
```
davehunt: We only need to test the CSS here as the JavaScript is still inline by default and tested in `test_resources`.

test_pytest_html.py
```diff
@@ -174,18 +174,34 @@ def test_create_report_path(self, testdir):
...
+                content = content.decode('utf-8')
+            assert content
+            assert content in html
+
     def test_resources(self, testdir):
         testdir.makepyfile('def test_pass(): pass')
         result, html = run(testdir)
         assert result.ret == 0
-        for resource in ['style.css', 'main.js']:
+        for resource in ['main.js']:
```
davehunt: No need for a loop when we only have one item.

test_pytest_html.py
```diff
@@ -174,18 +174,34 @@ def test_create_report_path(self, testdir):
...
             content = pkg_resources.resource_string(
                 'pytest_html', os.path.join('resources', resource))
             if PY3:
                 content = content.decode('utf-8')
             assert content
             assert content in html
 
+        regex_css_link = '<link href='
+        assert re.search(regex_css_link, html) is not None
```
davehunt: This is not explicit enough to be sure it's a style link.

pytest_html/plugin.py
```diff
@@ -41,11 +41,12 @@ def pytest_addoption(parser):
     group.addoption('--html', action='store', dest='htmlpath',
                     metavar='path', default=None,
                     help='create html report file at given path.')
-    group.addoption('--self-contained-html', action="store",
-                    dest='self_contained_html', default=False,
-                    help='If it is true, asserts that the HTML will not ' +
-                    'depend of external files, risk of breaking CSP. ' +
-                    'DEFAULT: False')
+    group.addoption('--self-contained-html', action='store_true',
+                    help='create a self-contained html file containing all ' +
```
davehunt: The `+` are not needed. See https://docs.python.org/2/reference/lexical_analysis.html#string-literal-concatenation

pytest_html/plugin.py
```diff
@@ -41,11 +41,12 @@ def pytest_addoption(parser):
...
+                    'necessary styles, scripts, and images - this means ' +
+                    'that the report may not render or function where CSP ' +
+                    'restrictions are in place (see' +
```
davehunt: Add a space after `(see`

pytest_html/plugin.py
```diff
@@ -334,14 +335,14 @@ def _save_report(self, report_content, self_contained_html):
         dir_name = os.path.dirname(self.logfile)
         if not os.path.exists(dir_name):
             os.makedirs(dir_name)
-        logfile = open(self.logfile, 'w', encoding='utf-8')
-        logfile.write(report_content)
+        with open(self.logfile, 'w', encoding='utf-8') as logfile:
```
davehunt: 👍 let's just call is `f` though instead of `logfile`.

pytest_html/plugin.py
```diff
@@ -334,14 +335,14 @@ def _save_report(self, report_content, self_contained_html):
...
+            logfile.write(report_content)
         logfile.close()
 
         if not self_contained_html:
-            style_path = dir_name + '/style.css'
-            style_file = open(style_path, 'w', encoding='utf-8')
-            style_file.write(self.style_css)
+            style_path = os.path.join(dir_name, 'style.css')
```
davehunt: 👍

pytest_html/plugin.py
```diff
@@ -334,14 +335,14 @@ def _save_report(self, report_content, self_contained_html):
...
+            with open(style_path, 'w', encoding='utf-8') as style_file:
```
davehunt: As above, let's call the file variable simply `f`.

test_pytest_html.py
```diff
@@ -176,30 +176,29 @@ def test_create_report_path(self, testdir):
...
-        regex_css_link = '<link href='
+
+        content = pkg_resources.resource_string(
+            'pytest_html', os.path.join('resources', 'main.js'))
+        if PY3:
+            content = content.decode('utf-8')
+        assert content
+        assert content in html
+
+        regex_css_link = '<link href="style.css" rel="stylesheet"'
```
davehunt: 👍 perfect

pytest_html/plugin.py
davehunt: In case you missed it, this is still outstanding.

pytest_html/plugin.py
davehunt: In case you missed it, this is still outstanding.

README.rst
```diff
@@ -87,6 +87,18 @@ conftest.py file:
               extra.append(pytest_html.extras.html('<div>Additional HTML</div>'))
           report.extra = extra
 
+Creating a single file to the project
```
davehunt: Great idea to add documentation for this. I would move it to before the 'Enhancing reports' section though as it is less advanced than that. I would also change the header to 'Creating a self-contained report'.

README.rst
```diff
@@ -87,6 +87,18 @@ conftest.py file:
...
+-------------------------------------
+
+In order to respect the `Content Security Policy (CSP) <https://developer.mozilla.org/docs/Web/Security/CSP>`_, the CSS file is created apart by default. But if it is necessary to you create a single file to the whole report, you may run the file by the following way:
```
davehunt: Try to keep the line length to 80 characters so the file is more readable.

davehunt: As we know there are more files going to be split out, let's not single out the CSS file in the documentation. We could say that 'several assets such as CSS and images' are stored separately by default'.

davehunt: We could give an example of where a self-contained report might be useful. Something like "You can alternatively create a self-contained report, which can be more convenient when sharing your results. This can be done in the following way:".

README.rst
```diff
@@ -87,6 +87,18 @@ conftest.py file:
...
+
+.. code-block:: bash
+
+    $ py.test --html=report.html --self-contained-html
+
+So, the CSS will be inline and the whole report will be created on a single file.
```
davehunt: I think this line is redundant.

pytest_html/plugin.py
```diff
@@ -320,13 +331,19 @@ def generate_summary_item(self):
...
-        if not os.path.exists(os.path.dirname(self.logfile)):
-            os.makedirs(os.path.dirname(self.logfile))
-        logfile = open(self.logfile, 'w', encoding='utf-8')
+    def _save_report(self, report_content, self_contained_html):
+        dir_name = os.path.dirname(self.logfile)
+        if not os.path.exists(dir_name):
+            os.makedirs(dir_name)
+        with open(self.logfile, 'w', encoding='utf-8') as f:
+            f.write(report_content)
+        f.close()
```
leitzler: When using "with ... as" in python, you don't need to close it afterwards, that is done automatically.

```python
>>> with open('test.txt', 'w') as f:
...     f.write('foo')
... 
3
>>> f.closed
True
```

pytest_html/plugin.py
```diff
@@ -320,13 +331,19 @@ def generate_summary_item(self):
...
 
-        logfile.write(report_content)
-        logfile.close()
+        if not self_contained_html:
+            style_path = os.path.join(dir_name, 'style.css')
+            with open(style_path, 'w', encoding='utf-8') as f:
+                f.write(self.style_css)
+            f.close()
```
leitzler: Unnecessary close here too.

README.rst
```diff
@@ -52,6 +52,21 @@ Then run your tests with:
 
   $ py.test --html=report.html
 
+
+Creating a a self-contained report
```
davehunt: Repeating 'a'

README.rst
```diff
@@ -87,6 +102,7 @@ conftest.py file:
               extra.append(pytest_html.extras.html('<div>Additional HTML</div>'))
           report.extra = extra
 
+
```
davehunt: Remove this extra newline.

pytest_html/plugin.py
```diff
@@ -41,6 +41,12 @@ def pytest_addoption(parser):
     group.addoption('--html', action='store', dest='htmlpath',
                     metavar='path', default=None,
                     help='create html report file at given path.')
+    group.addoption('--self-contained-html', action='store_true',
+                    help='create a self-contained html file containing all '
+                    'necessary styles, scripts, and images - this means '
+                    'that the report may not render or function where CSP '
+                    'restrictions are in place (see ' +
```
davehunt: Remove the trailing `+`

Issue #43, opened 2016-05-20
Error or failure information can be collapsed


hailin-xiao:
Hi,

Could you please add a new feature that user could collapse the error or failure log? it's a little bit hard to find the useful information once there're hundreds of errors or failures in the report. 
For better understanding, I would like to give an example: <div class='log'><details>balabala</details></div>
Thanks very much and looking forward for that.

Regards,

Hailin

davidedelvento:
@davehunt - I think this becomes even more important because of #44 (at least for some people).
I'll be happy to work on a rudimentary javascript implementation and make a PR, if no one is working on it already.

hailin-xiao:
@davidedelvento @davehunt 
Thanks for your quick response. Actually, we have made a temporary solution to satisfy our project. It's to make some changes on main.js. Please refer to the changes listed below. Any comments are appreciated. Thanks

Hailin

![changes](https://cloud.githubusercontent.com/assets/17964912/15527221/19569c24-226b-11e6-825e-3475c1efa1df.jpg)

davehunt:
Interesting, so this hides all log content by default and allows users to show the logs by clicking anywhere in the row? For my use case I prefer the log to be displayed by default, though I agree with @davidedelvento that I might prefer logs from passing tests to be hidden. Perhaps we could have a 'display log' link for these, and a 'hide log' link for any logs displayed?

We could also have 'show all logs' and 'hide all logs' links at the top of the page if users want to quickly toggle all logs. It then also sounds like we should have a preference for controlling if the logs are displayed initially, unless the 'hide all logs' does what we need there?

hailin-xiao:
@davehunt Thanks for your reply. It's just a temporary workaround for this  as we don't have enough time  to do this. I definitely agree with your solution. It could be great for user to choose display or hide the logs. But it would be better to let each log have such an option. We're looking forward to the solution be committed. Thanks a log.

Hailin

RibeiroAna:
[...]

PR #64, opened 2016-07-13
Expand / Collapse for extras-row

leitzler:
Proposal for #43

### DIFF
--- a/pytest_html/resources/main.js
+++ b/pytest_html/resources/main.js
@@ -38,8 +38,55 @@ function sort_column(elem) {
     sort_table(elem, key(colIndex));
 }
 
+function show_all_extras() {
+    find_all('.col-result').forEach(show_extras);
+}
+
+function hide_all_extras() {
+    find_all('.col-result').forEach(hide_extras);
+}
+
+function show_extras(colresult_elem) {
+    var extras = colresult_elem.parentNode.nextElementSibling;
+    var expandcollapse = colresult_elem.firstElementChild;
+    extras.classList.remove("collapsed");
+    expandcollapse.classList.remove("expander");
+    expandcollapse.classList.add("collapser");
+}
+
+function hide_extras(colresult_elem) {
+    var extras = colresult_elem.parentNode.nextElementSibling;
+    var expandcollapse = colresult_elem.firstElementChild;
+    extras.classList.add("collapsed");
+    expandcollapse.classList.remove("collapser");
+    expandcollapse.classList.add("expander");
+}
+
+function add_collapse() {
+    find_all('.col-result').forEach(function(elem) {
+        var extras = elem.parentNode.nextElementSibling;
+        var expandcollapse = document.createElement("span");
+        if (elem.innerHTML === 'Passed') {
+            extras.classList.add("collapsed");
+            expandcollapse.classList.add("expander");
+        } else {
+            expandcollapse.classList.add("collapser");
+        }
+        elem.appendChild(expandcollapse);
+
+        elem.addEventListener("click", function(event) {
+            if (event.currentTarget.parentNode.nextElementSibling.classList.contains("collapsed")) {
+                show_extras(event.currentTarget);
+            } else {
+                hide_extras(event.currentTarget);
+            }
+        });
+    })
+}
 addEventListener("DOMContentLoaded", function() {
     reset_sort_headers();
+    
+    add_collapse();
 
     sort_column(find('.initial-sort'));
 
--- a/pytest_html/resources/main.js
+++ b/pytest_html/resources/main.js
@@ -63,6 +63,14 @@ function hide_extras(colresult_elem) {
 }
 
 function add_collapse() {
+    // Add links for show/hide all
+    var resulttable = find('table#results-table');
+    var showhideall = document.createElement("p");
+    showhideall.innerHTML = '<a href="javascript:show_all_extras()">Show all details</a> / ' +
+                            '<a href="javascript:hide_all_extras()">Hide all detauls</a>';
+    resulttable.parentElement.insertBefore(showhideall, resulttable);
+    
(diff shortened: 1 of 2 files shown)


leitzler:
Looks like this now:
issue43

davehunt:
This looks great! Could we make the content to be "show details" and "hide details" to be a little more specific about what we're toggling? Also, could you make the mouse cursor change to indicate the show/hide is clickable?

Have you tested this with a large report? There have been some recent tweaks that reduced the amount of JavaScript on initial load, and I'm wondering if this might cause a regression.

leitzler:
Updated the pull request with "details" & cursor pointer.

I've not tested it yet with a large report, are you referring to large as "many tests" or "large output from a test case" (or both)?

Since I've only changed the .js & the .css the output will not be larger than the actual diff.

leitzler:
I added quite a huge test "test report" as a gist if you want to take a look at performance.
Note that it is almost 4MB :)

https://rawgit.com/leitzler/fdab4a0575a44394190fc7d701d7cd5f/raw/0f747201437d59fa1fb4c98e9ac5707c43f61b12/huge.html

davehunt:
That's a huge report! Perhaps create something with ~1000 results and compare page load times before/after your patch to see if it's regressed?

davidedelvento:
I tested this PR for my use case. Other than the missing of  "show all/hide all" links (which would be very convenient for me, in addition to the checkboxes that totally hide the entries in the table), the thing behaves very nicely. 
Hope it gets merged soon.
PS: I haven't found the time to look at the code yet, so can't comment on that.

leitzler:
@davehunt so, the new report will be 1808 bytes larger (uncompressed), that will not affect the performance notably :).

So for the load time, I ran a couple of page refreshes using Chrome 51 on Windows 10, and a report with 1000 passed tests, and 5 failed that produced a log with 5, 1000, 10000, 50000, and 100000 chars.

This is the result for the old one, 640798 bytes:

| Loading        | Scripting           | Rendering  |  Other  |  Idle  |
| -------------:| -------------:| -----:| ---:| ---:| ---:|
| 35ms      | 94ms | 294ms | 30ms | 65ms |
| 39ms      | 124ms | 353ms | 32ms | 109ms |
| 38ms      | 97ms | 335ms | 30ms | 61ms |
| 37ms      | 104ms | 341ms | 28ms | 62ms |
| avg: 37.25 | avg: 104.75 | avg: 330.75 | avg: 30.0 | 
New one, 642606 bytes:

| Loading        | Scripting           | Rendering  |  Other  |  Idle  |
| -------------:| -------------:| -----:| ---:| ---:| ---:|
| 36ms      | 114ms | 267ms | 29ms | 51ms |
| 37ms      | 104ms | 299ms | 26ms | 51ms |
| 38ms      | 109ms | 312ms | 27ms | 60ms |
| 39ms      | 117ms | 317ms | 27ms | 91ms |
| avg: 37.5 | avg: 111.0 | avg: 298.75 | avg: 27.25

I guess the scripting and rendering are the once that could have been affected by this patch. And I can't say that the difference is big enough to even be measurable using my method :).

davehunt:
@leitzler thanks for looking into that in such detail! I'm convinced. Any chance you could add the show/hide all links for @davidedelvento? Bonus points if we don't show these links if JavaScript is disabled. 😀

leitzler:
No problems, I use the plugin myself and missed this feature :). I added those links too. Hope it looks ok for you too @davidedelvento!

pytest_html/resources/main.js
```diff
@@ -38,8 +38,63 @@ function sort_column(elem) {
...
+    var showhideall = document.createElement("p");
+    showhideall.innerHTML = '<a href="javascript:show_all_extras()">Show all details</a> / ' +
+                            '<a href="javascript:hide_all_extras()">Hide all detauls</a>';
+    resulttable.parentElement.insertBefore(showhideall, resulttable);
+    
+    // Add show/hide link to each result
+    find_all('.col-result').forEach(function(elem) {
+        var extras = elem.parentNode.nextElementSibling;
+        var expandcollapse = document.createElement("span");
+        if (elem.innerHTML === 'Passed') {
```
davehunt: I think we should also collapse `XPassed` results by default.

davehunt:
This looks great - I just added a note about collapsing `XPassed` results by default too. Let's see what @davidedelvento thinks, but I'm looking forward to having this in the next release! 👍

leitzler:
Nice!
About XPassed, since the tests passed ***unexpectedly***, wouldn't it be a good idea to show the output default?

davehunt:
> About XPassed, since the tests passed unexpectedly, wouldn't it be a good idea to show the output default?

In most cases the output will be empty because the test passed, so showing an empty log has limited value. I'm also fine with leaving as it is.

leitzler:
Then I suggest we leave it like this for now. I know that there has been quite a few discussions about however XPassed is a failed state or not that resulted in the 'strict' parameter. Since it isn't obvious for all users, it's better to have it visible than hidden :).

davehunt:
That works for me - if I don't hear anything from @davidedelvento in the next day or so I'll go ahead and merge.

davidedelvento:
Sorry, busy weekend. 
I'll test it right now, and let you know my feedback in 30-60 minutes.
Thanks!

davidedelvento:
This is great. Looking forward to have that in a official release I can tell other people on my team to install (they can only use officially released sw)

davehunt:
Thanks @leitzler and @davidedelvento! I'm going to see if we can squeeze in a couple more merges before the next release, but it will be soon!

davehunt:
@leitzler you deleted that merged branch pretty quickly - is that a bot? I could do with something like that!

leitzler:
Manually, Github sent the mail notification pretty fast I guess :)

PR #68, opened 2016-07-20
Handling with Images

RibeiroAna:


pytest_html/plugin.py
```diff
@@ -55,7 +56,8 @@ def pytest_configure(config):
     htmlpath = config.option.htmlpath
     # prevent opening htmlpath on slave nodes (xdist)
     if htmlpath and not hasattr(config, 'slaveinput'):
-        config._html = HTMLReport(htmlpath)
+        contained_html = config.getoption('self_contained_html')
+        config._html = HTMLReport(htmlpath, contained_html=contained_html)
```
davehunt: I would call this variable and option `self_contained` as it's an attribute of the `HTMLReport`.

pytest_html/plugin.py
```diff
@@ -86,7 +88,7 @@ def data_uri(content, mime_type='text/plain', charset='utf-8'):
 
 class HTMLReport(object):
 
-    def __init__(self, logfile):
+    def __init__(self, logfile, **kwargs):
```
davehunt: Let's explicitly list the keyword arguments - it provides better documentation and usability of the class.

pytest_html/plugin.py
```diff
@@ -127,18 +131,36 @@ def __lt__(self, other):
...
-                        extra.get('content'))
-                additional_html.append(html.div(
-                    html.a(html.img(src=image), href="#"),
+                image_src = None
+                if self.contained_html:
+                    image_src = 'data:image/png;base64,{0}'.format(
+                            extra.get('content'))
+                    href = '#'
+                else:
+                    hash_key = self.test_id + self.outcome
```
davehunt: I'm not sure we need to include the outcome, but there may be multiple images, so we should distinguish them.

pytest_html/plugin.py
```diff
@@ -127,18 +131,36 @@ def __lt__(self, other):
...
+                    hash_generator = hashlib.md5()
+                    hash_generator.update(hash_key)
+                    image_file_name = '{0}.jpg'.format(hash_generator.digest)
```
davehunt: We currently only support PNG so we should use the appropriate file extension.

pytest_html/plugin.py
```diff
@@ -127,18 +131,36 @@ def __lt__(self, other):
...
+
+                    super_dir_name = os.path.dirname(self.logfile)
+                    dir_name = os.path.join(super_dir_name, 'assets')
+                    image_path = os.path.join(dir_name, image_file_name)
```
davehunt: I think we can reduce this down to:

```python
image_path = os.path.join(os.path.dirname(self.logfile), 'assets', image_file_name)
```

pytest_html/plugin.py
```diff
@@ -127,18 +131,36 @@ def __lt__(self, other):
...
+                    href = os.path.join('assets', image_file_name)
```
davehunt: Here you can use:

```python
href = os.path.relpath(image_path, os.path.dirname(self.logfile))
```

pytest_html/plugin.py
```diff
@@ -127,18 +131,36 @@ def __lt__(self, other):
...
-        def append_extra_html(self, extra, additional_html, links_html):
+        def append_extra_html(self, extra):
             href = None
             if extra.get('format') == extras.FORMAT_IMAGE:
-                href = '#'
-                image = 'data:image/png;base64,{0}'.format(
-                        extra.get('content'))
-                additional_html.append(html.div(
-                    html.a(html.img(src=image), href="#"),
+                image_src = None
```
davehunt: We don't need this line as there should always be a `src`

pytest_html/plugin.py
```diff
@@ -127,18 +131,36 @@ def __lt__(self, other):
...
+                    hash_generator.update(hash_key)
+                    image_file_name = '{0}.jpg'.format(hash_generator.digest)
+
+                    super_dir_name = os.path.dirname(self.logfile)
+                    dir_name = os.path.join(super_dir_name, 'assets')
+                    image_path = os.path.join(dir_name, image_file_name)
+                    href = os.path.join('assets', image_file_name)
+                    image_src = href
+
+                    with open(image_path, 'w') as fh:
```
davehunt: We should probably create the assests directory in case it doesn't already exist. We do this in `_save_report` already, but that's too late for writing the images to it.

test_pytest_html.py
```diff
@@ -287,7 +288,7 @@ def pytest_runtest_makereport(item, call):
                     report.extra = [extras.image('{0}')]
         """.format(content))
         testdir.makepyfile('def test_pass(): pass')
-        result, html = run(testdir)
+        result, html = run(testdir,  'report.html', '--self-contained-html')
```
davehunt: Nit: Too many spaces between arguments.

davehunt:
Great work! Just a few comments to address and a new test to write. 🎉

pytest_html/plugin.py
RibeiroAna: @davehunt I think didn't understand very well, are you suggesting to use single arguments instead of kwargs?  Or what?

RibeiroAna: @davehunt I think didn't understand very well, are you suggesting to use single arguments instead of kwargs?  Or what?

pytest_html/plugin.py
RibeiroAna: I didn't get what you want to say here as well... What are you suggesting?

pytest_html/plugin.py
RibeiroAna: I remained with dir_name because we need it to create the folder also.

pytest_html/plugin.py
davehunt: Yes, avoid using `**kwargs` unless you're unable to predict the arguments being passed. In our case there's a new optional argument named `self_contained` and the value should default to `False`. If we need additional arguments later we can add them.

pytest_html/plugin.py
davehunt: We want the hash to be unique, so the key should also be unique. I'm not sure what benefit adding the `outcome` to the key provides? The report may include multiple images, and we'd also want each image to have a unique key. I would suggest a combination of `test_id` and the index of the extra that we're currently handling. We may also need to consider reruns, so perhaps if it's available the index of the current rerun.

pytest_html/plugin.py
davehunt: You should be able to run `os.makedirs` with the full path to the image or if not, you could use `os.path.dirname`.

pytest_html/plugin.py
RibeiroAna: The only thing that the `outcome` may help is with rerun, because the the test which failed at first may have a different picture, but  I don't see a way to get the index of the rerun. Do you have any idea?

RibeiroAna: I got an idea, can be the number of the tests, like there will be 5 tests passed or 3 reruns, so we can pass this number as a parameter

davehunt: The report should have a `rerun` attribute with an integer indicating the run number. You could use something like `getattr` with a default of 0 if the attribute is not present.

RibeiroAna: The way I thought is easier, and I think it works by the same way (it wont be 2 third reruns, etc)

pytest_html/plugin.py
```diff
@@ -112,8 +113,10 @@ def __init__(self, outcome, report, contained_html, logfile):
             self.contained_html = contained_html
             self.logfile = logfile
 
+            extra_index = 1
             for extra in getattr(report, 'extra', []):
```
davehunt: If you use [`enumerate`](https://docs.python.org/2.7/library/functions.html#enumerate) then you will be able to access the index.

pytest_html/plugin.py
```diff
@@ -131,24 +134,27 @@ def __lt__(self, other):
...
                     href = '#'
                 else:
-                    hash_key = self.test_id + self.outcome
+                    hash_key = self.test_id + extra_index + test_index
                     hash_generator = hashlib.md5()
                     hash_generator.update(hash_key)
-                    image_file_name = '{0}.jpg'.format(hash_generator.digest)
+                    image_file_name = '{0}.png'.format(hash_generator.digest)
+
+                    dir_name = os.path.join(os.path.dirname(self.logfile),
```
davehunt: I would still prefer to just have the image path and use that for creating any necessary directories, but I'll leave it up to you.

pytest_html/plugin.py
```diff
@@ -112,8 +113,10 @@ def __init__(self, outcome, report, contained_html, logfile):
             self.contained_html = contained_html
             self.logfile = logfile
 
+            extra_index = 1
             for extra in getattr(report, 'extra', []):
-                self.append_extra_html(extra)
+                self.append_extra_html(extra, extra_index, test_index)
```
davehunt: I understand why you went with the number of tests with a specific outcome, but I think using `report.rerun` would be a better solution. Your solution will mean a different hash will be generated if the tests run in a different order, or if a test has a different outcome. By using the run number for the individual test we'll be able to rewrite files in the assets directory on subsequent runs, causing less junk to build up. It also means we can avoid passing the additional integer for every `TestResult` object, as we already have the `report` object.

pytest_html/plugin.py
```diff
@@ -97,11 +97,12 @@ def __init__(self, logfile, **kwargs):
         self.passed = self.skipped = 0
         self.xfailed = self.xpassed = 0
         self.rerun = 0
-        self.contained_html = kwargs['contained_html']
+        self.contained_html = self_contained
```
davehunt: Please rename this to `self_contained` throughout.

pytest_html/plugin.py
```diff
@@ -261,7 +267,7 @@ def _generate_report(self, session):
         css_href = os.path.join('assets', 'style.css')
         html_css = html.link(href=css_href, rel='stylesheet',
                              type='text/css')
-        if session.config.getoption('self_contained_html'):
+        if session.config.getoption('self_contained'):
```
davehunt: We shouldn't need to get this from `session.config` now.

pytest_html/plugin.py
```diff
@@ -368,7 +374,7 @@ def generate_summary_item(self):
             unicode_doc = unicode_doc.decode('utf-8')
         return unicode_doc
 
-    def _save_report(self, report_content, self_contained_html):
+    def _save_report(self, report_content, self_contained):
```
davehunt: We shouldn't need to pass this in anymore as it's available from `HTMLReport`

pytest_html/plugin.py
```diff
@@ -397,8 +403,8 @@ def pytest_sessionstart(self, session):
 
     def pytest_sessionfinish(self, session):
         report_content = self._generate_report(session)
-        self_contained_html = session.config.getoption('self_contained_html')
-        self._save_report(report_content, self_contained_html)
+        self_contained = session.config.getoption('self_contained')
```
davehunt: Again, we shouldn't need to take this from `session.config` anymore.

test_pytest_html.py
```diff
@@ -198,7 +199,8 @@ def test_resources(self, testdir):
         assert content
         assert content in html
 
-        regex_css_link = '<link href="assets/style.css" rel="stylesheet"'
+        href = os.path.join('assets', 'style.css')
```
davehunt: This isn't a file path, it's a relative URL, so we don't want to use `os.path.join` here.

test_pytest_html.py
```diff
@@ -287,12 +289,36 @@ def pytest_runtest_makereport(item, call):
...
+                    report.extra = [extras.image('{0}')]
+        """.format(content))
+        testdir.makepyfile('def test_pass(): pass')
+        result, html = run(testdir)
+        hash_key = 'test_extra_image_separeted0::test.py11'
+        hash_generator = hashlib.md5()
+        hash_generator.update(hash_key)
+        assert result.ret == 0
+        a_img = ('<a class="image" href="{0}" '
+                'target="_blank">Image</a>'.format(src))
```
davehunt: You're using `src` before it's defined.

RibeiroAna:
Just saying, it is not ready yet for more reviews, I still need to do the rerun number.

RibeiroAna:
Now you may review, @davehunt

pytest_html/plugin.py
```diff
@@ -5,6 +5,7 @@
 from __future__ import absolute_import
 
 from base64 import b64encode
+from base64 import b64decode
```
davehunt: You can import both of these on the same line, by seperating them with a comma:

```python
from base64 import b64encode, b64decode
```

pytest_html/plugin.py
```diff
@@ -95,23 +98,27 @@ def __init__(self, logfile):
...
             self.time = getattr(report, 'duration', 0.0)
             self.outcome = outcome
             self.additional_html = []
             self.links_html = []
+            self.self_contained = self_contained
+            self.logfile = logfile
 
-            self.append_log_html(report, self.additional_html)
+            for extra_index, extra in enumerate(getattr(report, 'extra', []),
+                                                start=1):
```
davehunt: why `start=1`?

pytest_html/plugin.py
```diff
@@ -95,23 +98,27 @@ def __init__(self, logfile):
         self.passed = self.skipped = 0
         self.xfailed = self.xpassed = 0
         self.rerun = 0
+        self.self_contained = self_contained
 
     class TestResult:
 
-        def __init__(self, outcome, report):
+        def __init__(self, outcome, report, self_contained,
+                     logfile, test_index):
```
davehunt: I still don't think we need `test_index` as we already have the report. Can you try using `getattr(report, 'reruns', 0)` in the call to `append_extra_html`?

pytest_html/plugin.py
```diff
@@ -127,18 +134,41 @@ def __lt__(self, other):
...
-                additional_html.append(html.div(
-                    html.a(html.img(src=image), href="#"),
+                if self.self_contained:
+                    image_src = 'data:image/png;base64,{0}'.format(
+                            extra.get('content'))
+                    href = '#'
+                else:
+                    hash_key = "{0}{1}{2}".format(
+                               self.test_id, str(extra_index),
+                               str(test_index)).encode('utf-8')
```
davehunt: If we're just concatenating these, try using `''.join([self.test_id, extra_index, test_index])`

pytest_html/plugin.py
```diff
@@ -127,18 +134,41 @@ def __lt__(self, other):
...
+                    hash_generator = hashlib.md5()
+                    hash_generator.update(hash_key)
+                    image_file_name = '{0}.png'.format(hash_generator
+                                                       .hexdigest())
+                    dir_name = os.path.join(os.path.dirname(self.logfile),
+                                            'assets')
+                    if not os.path.exists(dir_name):
+                        os.makedirs(dir_name)
+
+                    image_path = os.path.join(dir_name, image_file_name)
```
davehunt: We still don't really need the intermediate `dir_name` variable:

```python
image_path = os.path.join(os.path.dirname(self.logfile), 'assets', image_file_name)
if not os.path.exists(os.path.dirname(image_path)):
    os.makedirs(os.path.dirname(image_path))
```

pytest_html/plugin.py
```diff
@@ -344,14 +376,17 @@ def generate_summary_item(self):
...
         return unicode_doc
 
-    def _save_report(self, report_content, self_contained_html):
+    def _save_report(self, report_content):
         dir_name = os.path.dirname(self.logfile)
+        assets_dir = os.path.join(dir_name, 'assets')
         if not os.path.exists(dir_name):
             os.makedirs(dir_name)
+        if not os.path.exists(assets_dir):
+            os.makedirs(assets_dir)
```
davehunt: As this is beneath `dir_name` we only need to call `makedirs` on this, which will make the parent directory if needed.

davehunt: We don't want to create the `assets` directory when `self.self_contained` is `True`.

test_pytest_html.py
```diff
@@ -287,12 +287,72 @@ def pytest_runtest_makereport(item, call):
...
+                    'test_pass1-1').encode('utf-8')
+        hash_generator = hashlib.md5()
+        hash_generator.update(hash_key)
+        assert result.ret == 0
+        src = os.path.join('assets',
+                           '{0}.png'.format(hash_generator.hexdigest()))
+        a_img = ('<a class="image" href="{0}" '
+                 'target="_blank">Image</a>'.format(src))
+        assert a_img in html
+        assert '<a href="{0}"><img src="{1}"/></a>'.format(src, src) in html
```
davehunt: We should check that the generated images exist on the disk.

RibeiroAna: How we do that?

davehunt: https://docs.python.org/2/library/os.path.html#os.path.exists

pytest_html/plugin.py
RibeiroAna: Since the reruns starts on 0, I'll let the default as -1, so we can know which was the first rerun and the failed test

davehunt: Ah, so the first rerun is `0`? Makes sense to use `-1` as the default then.

pytest_html/plugin.py
```diff
@@ -114,8 +112,12 @@ def __init__(self, outcome, report, self_contained,
             self.self_contained = self_contained
             self.logfile = logfile
 
+            test_index = -1
+            if outcome == 'Rerun':
+                test_index = report.rerun
+
             for extra_index, extra in enumerate(getattr(report, 'extra', []),
-                                                start=1):
+                                                start=0):
```
davehunt: Isn't this zero indexed by default? I don't think we need to specify the `start` argument at all.

pytest_html/plugin.py
```diff
@@ -95,23 +97,30 @@ def __init__(self, logfile):
...
             self.outcome = outcome
             self.additional_html = []
             self.links_html = []
+            self.self_contained = self_contained
+            self.logfile = logfile
 
-            self.append_log_html(report, self.additional_html)
+            test_index = -1
+            if outcome == 'Rerun':
+                test_index = report.rerun
```
davehunt: okay, I think I understand this now. what about:

```python
test_index = hasattr(report, 'rerun') and report.rerun or -1
```

pytest_html/plugin.py
```diff
@@ -127,18 +136,44 @@ def __lt__(self, other):
...
+                    href = os.path.relpath(image_path,
+                                           os.path.dirname(self.logfile))
+                    image_src = href
+
+                    with open(image_path, 'w') as f:
+                        content = extra.get('content')
+                        if PY3:
+                            content = bytearray()
+                            content.extend(extra.get('content')
+                                           .encode('utf-8'))
```
davehunt: It looks like `bytearray` can be [initialized](https://docs.python.org/3/library/functions.html#bytearray) with `source` and `encoding`.

pytest_html/plugin.py
```diff
@@ -344,14 +381,21 @@ def generate_summary_item(self):
             unicode_doc = unicode_doc.decode('utf-8')
         return unicode_doc
 
-    def _save_report(self, report_content, self_contained_html):
+    def _save_report(self, report_content):
         dir_name = os.path.dirname(self.logfile)
-        if not os.path.exists(dir_name):
-            os.makedirs(dir_name)
+        assets_dir = os.path.join(dir_name, 'assets')
```
davehunt: We hard-code this in a few places - do you think it would make sense to add it to the `HTMLReport` as an attribute? We could then just use `self.assets_path` throughout.

pytest_html/plugin.py
```diff
@@ -344,14 +381,21 @@ def generate_summary_item(self):
...
+
+        if self.self_contained:
+            if not os.path.exists(dir_name):
+                os.makedirs(dir_name)
+        else:
+            if not os.path.exists(assets_dir):
+                os.makedirs(assets_dir)
```
davehunt: Let's optimize this further:

```python
if not os.path.exists(dir_name):
    os.makedirs(dir_name)
if self.self_contained and not os.path.exists(assets_dir):
    os.makedirs(assets_dir)
```

pytest_html/plugin.py
```diff
@@ -235,9 +271,10 @@ def _generate_report(self, session):
         if PY3:
             self.style_css = self.style_css.decode('utf-8')
 
-        html_css = html.link(href='style.css', rel='stylesheet',
+        css_href = os.path.join('assets', 'style.css')
```
davehunt: This needs to be joined by `/` and not the system path separator, otherwise it will fail on Windows.

pytest_html/plugin.py
RibeiroAna: Hi @davehunt, it doesn't work, because the rerun counting starts with 0...

test_pytest_html.py
```diff
@@ -287,12 +287,74 @@ def pytest_runtest_makereport(item, call):
...
         """.format(content))
         testdir.makepyfile('def test_pass(): pass')
-        result, html = run(testdir)
+        result, html = run(testdir, 'report.html', '--self-contained-html')
         assert result.ret == 0
         assert '<a class="image" href="#" target="_blank">Image</a>' in html
         src = 'data:image/png;base64,{0}'.format(content)
         assert '<a href="#"><img src="{0}"/></a>'.format(src) in html
 
+    def test_extra_image_separeted(self, testdir):
```
davehunt: Typo: seperated

test_pytest_html.py
```diff
@@ -287,12 +287,74 @@ def pytest_runtest_makereport(item, call):
...
+                outcome = yield
+                report = outcome.get_result()
+                if report.when == 'call':
+                    from pytest_html import extras
+                    report.extra = [extras.image('{0}')]
+        """.format(content))
+        testdir.makepyfile('def test_pass(): pass')
+        result, html = run(testdir)
+        hash_key = ('test_extra_image_separeted.py::'
+                    'test_pass0-1').encode('utf-8')
```
davehunt: I'm having second thoughts on the `-1` after seeing it here. Functionally it's fine but it looks ugly to me. 😝 I suppose we could increment the reruns so they're never `0`. I'll let you decide on that. 😀

test_pytest_html.py
```diff
@@ -287,12 +287,74 @@ def pytest_runtest_makereport(item, call):
...
+            def pytest_runtest_makereport(item, call):
+                outcome = yield
+                report = outcome.get_result()
+                if report.when == 'call':
+                    from pytest_html import extras
+                    report.extra = [extras.image('{0}')]
+        """.format(content))
+        testdir.makepyfile('def test_pass(): pass')
+        result, html = run(testdir)
+        hash_key = ('test_extra_image_separeted.py::'
```
davehunt: You should be able to get the test name using something like `request.node.name` to avoid having to hard-code it here.

test_pytest_html.py
```diff
@@ -287,12 +287,74 @@ def pytest_runtest_makereport(item, call):
...
+                    'test_pass0-1').encode('utf-8')
+        hash_generator = hashlib.md5()
+        hash_generator.update(hash_key)
+        assert result.ret == 0
+        src = os.path.join('assets',
```
davehunt: This should be explicitly joined with `/` and not the system path separator.

pytest_html/plugin.py
```diff
@@ -127,18 +136,44 @@ def __lt__(self, other):
...
+                    hash_generator.update(hash_key)
+                    image_file_name = '{0}.png'.format(hash_generator
+                                                       .hexdigest())
+                    image_path = os.path.join(os.path.dirname(self.logfile),
+                                              'assets', image_file_name)
+                    if not os.path.exists(os.path.dirname(image_path)):
+                        os.makedirs(os.path.dirname(image_path))
+
+                    href = os.path.relpath(image_path,
+                                           os.path.dirname(self.logfile))
```
davehunt: I just noticed that we need this to be separated by `/` and not the system path separator, otherwise it will fail on Windows.

test_pytest_html.py
davehunt: You can repeat `{0}` rather than pass the same variable into the formatter twice.

test_pytest_html.py
```diff
@@ -287,12 +287,74 @@ def pytest_runtest_makereport(item, call):
...
+                    report.extra = [extras.image('{0}')]
+        """.format(content))
+        testdir.makepyfile("""
+            import pytest
+            @pytest.mark.flaky(reruns=2)
+            def test_fail():
+                assert False""")
+        result, html = run(testdir)
+
+        for i in [0, 1]:
```
davehunt: If we rerun twice we should have three images. Either rerun just once or check all three. You can use `range(3)` to generate `[0, 1, 2]` or `range(-1, 2)` if you want to generate `[-1, 0, 1]`.

davehunt: Can you rename `i` to `rerun` to make it a little more clear?

pytest_html/plugin.py
davehunt: That shouldn't matter, if it's 0 then it would return 0, if it's not present it would return -1

davehunt: See my later comment about possibly even incrementing it to avoid using a negative number:

```python
test_index = hasattr(report, 'rerun') and report.rerun + 1 or 0
```

pytest_html/plugin.py
RibeiroAna: Well, now I saw that in fact it is used in many places, but in different classes, so it would have to be an argument of TestResult and HTMLReport, so it would have no difference.

test_pytest_html.py
RibeiroAna: I didn't understand very well, where is that object request that I can access in that function?... @davehunt

test_pytest_html.py
```diff
@@ -287,12 +287,73 @@ def pytest_runtest_makereport(item, call):
...
+                    report.extra = [extras.image('{0}')]
+        """.format(content))
+        testdir.makepyfile("""
+            import pytest
+            @pytest.mark.flaky(reruns=2)
+            def test_fail():
+                assert False""")
+        result, html = run(testdir)
+
+        for i in range(1, 3):
```
davehunt: There should be three results here (two reruns and a failure) so we can just use `range(3)`

test_pytest_html.py
```diff
@@ -287,12 +287,73 @@ def pytest_runtest_makereport(item, call):
...
+        """.format(content))
+        testdir.makepyfile('def test_pass(): pass')
+        result, html = run(testdir)
+        hash_key = ('test_extra_image_seperated.py::'
+                    'test_pass01').encode('utf-8')
+        hash_generator = hashlib.md5()
+        hash_generator.update(hash_key)
+        assert result.ret == 0
+        src = '{0}/{1}'.format('assets', '{0}.png'.
+                               format(hash_generator.hexdigest()))
```
davehunt: This is fine, but so you know you could also do:

```python
'/'.join(['assets', '{0}.png'.format(hash_generator.hexdigest())])
```

This also applies to a few other places, but I'm fine with leaving it as you have it here.

test_pytest_html.py
davehunt: It's a pytest fixture named `request`

pytest_html/plugin.py
davehunt: Good point.

test_pytest_html.py
RibeiroAna: Actually, it should be range (1, 4), because it starts on 1, since the first rerun is 1 (0 + 1), the second is 2 (1 + 1) and the test failed is 3 (2 + 1)

test_pytest_html.py
RibeiroAna: Since both are good, I think I'll let the original way, because there are many things like that, and I can forget one or two, and have a confused two styles on the code. I think changing that needs lot of time and attention.

Issue #66, opened 2016-07-20
Do not display filter checkboxes when JavaScript is unavailable


davehunt:
When JavaScript is disabled (such as via CSP) we should not display the filter checkboxes. The functionality is dependent on JavaScript, and without it they are just a distraction.

PR #70, opened 2016-07-27
No checkboxes when JS is disabled

RibeiroAna:
Related to #66

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -294,10 +294,11 @@ class HTMLReport(object):
                    class_='filter',
                    hidden=True)]
 
-        for outcome in outcomes:
-            summary.append(' -')
+        for i, outcome in enumerate(outcomes, start=1):
             summary.append(outcome.checkbox)
             summary.append(outcome.summary_item)
+            if i < len(outcomes):
+                summary.append(', ')
 
         results = [html.h2('Results'), html.table([html.thead(
             html.tr([
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -260,7 +260,8 @@ class HTMLReport(object):
 
             def generate_checkbox(self):
                 checkbox_kwargs = {'data-test-result':
-                                   self.test_result.lower()}
+                                   self.test_result.lower(),
+                                   'hidden': 'true'}
                 if self.total == 0:
                     checkbox_kwargs['disabled'] = 'true'
 
@@ -271,9 +272,10 @@ class HTMLReport(object):
                                            **checkbox_kwargs)
 
             def generate_summary_item(self):
-                self.summary_item = html.span('{0} {1}'.
+                self.summary_item = [html.span('{0} {1}'.
                                               format(self.total, self.label),
-                                              class_=self.class_html)
+                                              class_=self.class_html),
+                                     html.span('; ')]
 
         outcomes = [Outcome('passed', self.passed),
                     Outcome('skipped', self.skipped),
@@ -288,7 +290,9 @@ class HTMLReport(object):
         summary = [html.h2('Summary'), html.p(
             '{0} tests ran in {1:.2f} seconds. '.format(
                 numtests, suite_time_delta)),
-            html.p('(Un)check the boxes to filter the results.')]
+            html.p('(Un)check the boxes to filter the results.',
+                   id='filter_checkbox_text',
+                   hidden='true')]
 
         for outcome in outcomes:
             summary.append(outcome.checkbox)
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -260,8 +260,7 @@ class HTMLReport(object):
 
             def generate_checkbox(self):
                 checkbox_kwargs = {'data-test-result':
-                                   self.test_result.lower(),
-                                   'hidden': 'true'}
+                                   self.test_result.lower()}
                 if self.total == 0:
                     checkbox_kwargs['disabled'] = 'true'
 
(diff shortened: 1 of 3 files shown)


pytest_html/plugin.py
```diff
@@ -271,9 +272,10 @@ def generate_checkbox(self):
                                            **checkbox_kwargs)
 
             def generate_summary_item(self):
-                self.summary_item = html.span('{0} {1}'.
-                                              format(self.total, self.label),
-                                              class_=self.class_html)
+                self.summary_item = [html.span('{0} {1}'.
+                                               format(self.total, self.label),
+                                               class_=self.class_html),
+                                     html.span('; ')]
```
RibeiroAna: I added it, just because I think it looks better like that, when there isn't nothing to separate, I get confused when a report starts and other one finishes

pytest_html/plugin.py
```diff
@@ -260,7 +260,8 @@ def __init__(self, outcome, total=0, label=None,
 
             def generate_checkbox(self):
                 checkbox_kwargs = {'data-test-result':
-                                   self.test_result.lower()}
+                                   self.test_result.lower(),
+                                   'hidden': 'true'}
```
davehunt: This doesn't need to be in `checkbox_kwargs` as it can just be `hidden='true'` in the `self.checkbox` assignment.

pytest_html/plugin.py
davehunt: I agree it looks better with a separator, as this is what we had before the filters were introduced. Can you move this to the `for outcome in outcomes` line though, as we already have a separator there. You can also use `enumerate` and compare the index to the length of `outcomes` to avoid adding the separator for the last item.

pytest_html/plugin.py
```diff
@@ -288,7 +290,9 @@ def generate_summary_item(self):
         summary = [html.h2('Summary'), html.p(
             '{0} tests ran in {1:.2f} seconds. '.format(
                 numtests, suite_time_delta)),
-            html.p('(Un)check the boxes to filter the results.')]
+            html.p('(Un)check the boxes to filter the results.',
+                   id='filter_checkbox_text',
```
davehunt: We can change the checkboxes to use `class_='filter'` or similar, and reuse the class here. Then in the JavaScript we can just find elements with that class name and show them all. There would be no need to add the `id` here then.

pytest_html/resources/main.js
```diff
@@ -62,14 +62,23 @@ function hide_extras(colresult_elem) {
     expandcollapse.classList.add("expander");
 }
 
+function add_filter_boxes() {
+    var checkboxes = document.getElementsByName('filter_checkbox');
+    for (var i = 0; i < checkboxes.length; i++)
+      checkboxes[i].hidden = false;
+
+    var checkbox_text = document.getElementById('filter_checkbox_text');
+    checkbox_text.hidden = false;
```
davehunt: As mentioned above, by using a common class name we could then do `document.getElementsByClassName('filter')`. This function could then be renamed to `show_filters` and used for toggling visibility of all filter related elements that might be added in the future.

davehunt:
Works great - just a few comments to address.

pytest_html/plugin.py
```diff
@@ -288,12 +290,14 @@ def generate_summary_item(self):
         summary = [html.h2('Summary'), html.p(
             '{0} tests ran in {1:.2f} seconds. '.format(
                 numtests, suite_time_delta)),
-            html.p('(Un)check the boxes to filter the results.')]
+            html.p('(Un)check the boxes to filter the results.',
+                   class_='filter',
+                   hidden=True)]
```
davehunt: Use `hidden='true'` as on line 272.

pytest_html/plugin.py
```diff
@@ -268,12 +268,14 @@ def generate_checkbox(self):
...
+                                           hidden='true',
                                            **checkbox_kwargs)
 
             def generate_summary_item(self):
-                self.summary_item = html.span('{0} {1}'.
-                                              format(self.total, self.label),
-                                              class_=self.class_html)
+                self.summary_item = [html.span('{0} {1}'.
+                                               format(self.total, self.label),
+                                               class_=self.class_html)]
```
davehunt: You don't need to make this a list.

pytest_html/plugin.py
```diff
@@ -288,12 +290,14 @@ def generate_summary_item(self):
...
-            html.p('(Un)check the boxes to filter the results.')]
+            html.p('(Un)check the boxes to filter the results.',
+                   class_='filter',
+                   hidden=True)]
 
         for outcome in outcomes:
+            summary.append(' -')
             summary.append(outcome.checkbox)
             summary.append(outcome.summary_item)
-            summary.append(' ')
```
davehunt: This adds the separator at the beginning too, and misses to include a space after the dash. I would suggest the following:

```python
for i, outcome in enumerate(outcomes, start=1):
    summary.append(outcome.checkbox)
    summary.append(outcome.summary_item)
    if i < len(outcomes):
        summary.append(', ')
```

pytest_html/resources/main.js
```diff
@@ -62,14 +62,20 @@ function hide_extras(colresult_elem) {
     expandcollapse.classList.add("expander");
 }
 
+function show_filters() {
+    var filters_itens = document.getElementsByClassName('filter');
```
davehunt: Typo in variable name.

pytest_html/resources/main.js
```diff
@@ -62,14 +62,20 @@ function hide_extras(colresult_elem) {
...
+      filters_itens[i].hidden = false;
+}
+
 function add_collapse() {
     // Add links for show/hide all
     var resulttable = find('table#results-table');
     var showhideall = document.createElement("p");
     showhideall.innerHTML = '<a href="javascript:show_all_extras()">Show all details</a> / ' +
-                            '<a href="javascript:hide_all_extras()">Hide all detauls</a>';
+                            '<a href="javascript:hide_all_extras()">Hide all details</a>';
```
davehunt: 👍

pytest_html/resources/main.js
```diff
@@ -62,14 +62,20 @@ function hide_extras(colresult_elem) {
     expandcollapse.classList.add("expander");
 }
 
+function show_filters() {
+    var filter_itens = document.getElementsByClassName('filter');
```
davehunt: You fixed one typo in this, but not the one I spotted 😉 (itens/items)

davehunt: Once this is fixed, it's good to merge. 👍

PR #71, opened 2016-07-27
Removed rerun outcome unless the plugin is installed

RibeiroAna:


pytest_html/plugin.py
```diff
@@ -88,15 +90,17 @@ def data_uri(content, mime_type='text/plain', charset='utf-8'):
...
         self.logfile = os.path.abspath(logfile)
         self.test_logs = []
         self.results = []
         self.errors = self.failed = 0
         self.passed = self.skipped = 0
         self.xfailed = self.xpassed = 0
-        self.rerun = 0
+        self.has_rerun = has_rerun
+        if self.has_rerun:
+            self.rerun = 0
```
davehunt: How about:

```python
self.rerun = self.has_rerun and 0 or None
```

That way you just have one attribute, and you can later do `if self.rerun is not None`.

PR #72, opened 2016-07-29
Add files to external sources

RibeiroAna:


pytest_html/plugin.py
```diff
@@ -135,6 +135,26 @@ def __lt__(self, other):
...
+            if not os.path.exists(os.path.dirname(image_path)):
+                os.makedirs(os.path.dirname(image_path))
+
+            relative_path = '{0}/{1}'.format('assets', image_file_name)
+
+            with open(image_path, 'w') as f:
+                f.write(content)
+            return relative_path
+
+
```
RibeiroAna: @davehunt there are some flake8 problems and it is missing some tests, but I hope you can take a look before you go  =)

test_pytest_html.py
```diff
@@ -261,6 +255,31 @@ def pytest_runtest_makereport(item, call):
...
+        hash_key = ('test_extra_text_separeted.py::'
+                    'test_pass01').encode('utf-8')
+        hash_generator = hashlib.md5()
+        hash_generator.update(hash_key)
+        href = '{0}/{1}'.format('assets', '{0}.txt'.
+                               format(hash_generator.hexdigest()))
+        link = '<a class="text" href="{0}" target="_blank">Text</a>'.format(
+            href)
+        assert link in html
+
```
RibeiroAna: @davehunt I intend to do one test of text to rerun, and two new ones to JSON (separated "normal" and separated  with rerun (and now I realized a typo in "separeted")

pytest_html/plugin.py
```diff
@@ -135,6 +135,26 @@ def __lt__(self, other):
                      'XPassed', 'Skipped', 'Passed')
             return order.index(self.outcome) < order.index(other.outcome)
 
+        def save_external_src(self, content, extra_index,
```
davehunt: A better name might be `create_asset`

pytest_html/plugin.py
```diff
@@ -135,6 +135,26 @@ def __lt__(self, other):
...
+                              test_index, file_extension):
+            hash_key = ''.join([self.test_id, str(extra_index),
+                               str(test_index)]).encode('utf-8')
+            hash_generator = hashlib.md5()
+            hash_generator.update(hash_key)
+            image_file_name = '{0}.{1}'.format(hash_generator.hexdigest(),
```
davehunt: `asset_file_name` would be a better name here.

pytest_html/plugin.py
```diff
@@ -135,6 +135,26 @@ def __lt__(self, other):
...
+                                               file_extension)
+            image_path = os.path.join(os.path.dirname(self.logfile),
```
davehunt: `asset_path`

test_pytest_html.py
```diff
@@ -116,12 +116,6 @@ def test_example():
         assert result.ret
         assert_results(html, passed=0, failed=1, rerun=5)
 
-    def test_no_rerun(self, testdir):
-        testdir.makepyfile('def test_pass(): pass')
-        result, html = run(testdir, 'report.html', '-p', 'no:rerunfailures')
-        assert result.ret == 0
-        assert re.search('data-test-result="rerun"', html) is None
-
```
davehunt: I'm guessing this was a mistake?

test_pytest_html.py
```diff
@@ -261,6 +255,31 @@ def pytest_runtest_makereport(item, call):
...
+                outcome = yield
+                report = outcome.get_result()
+                if report.when == 'call':
+                    from pytest_html import extras
+                    report.extra = [extras.text('{0}')]
+        """.format(content))
+        testdir.makepyfile('def test_pass(): pass')
+        result, html = run(testdir)
+        assert result.ret == 0
+        hash_key = ('test_extra_text_separeted.py::'
```
davehunt: Can we get the name from `request` fixture? I recall mentioning this on another pull but perhaps we didn't do it there..

test_pytest_html.py
```diff
@@ -299,7 +318,7 @@ def pytest_runtest_makereport(item, call):
         src = 'data:image/png;base64,{0}'.format(content)
         assert '<a href="#"><img src="{0}"/></a>'.format(src) in html
 
-    def test_extra_image_seperated(self, testdir):
+    def test_extra_image_separeted(self, testdir):
```
davehunt: It's 'separated' :)

test_pytest_html.py
davehunt: It feels like a lot of duplication. I wonder if we can use a parametrized fixture with extra type, and extension as arguments?

test_pytest_html.py
RibeiroAna: How we can access the request from the test? And I think it is a good way to assert that the tests are being generated the way we want: the test id, plus the id of the test and the id of the fixture. I think if we just get it from the request, we could not assert that it is being generated by the key we want.

test_pytest_html.py
```diff
@@ -261,27 +261,8 @@ def pytest_runtest_makereport(item, call):
...
-        """.format(content))
-        testdir.makepyfile('def test_pass(): pass')
-        result, html = run(testdir)
-        assert result.ret == 0
-        link = '<a class="url" href="{0}" target="_blank">URL</a>'.format(
-            content)
-        assert link in html
-
-    def test_extra_image(self, testdir):
-        content = str(random.random())
```
RibeiroAna: That was another mistake, I'll correct that tomorrow

test_pytest_html.py
davehunt: request is a fixture provided by pytest. I'm only suggesting retrieving the test method name, which you can then combine with the other variables to create your test hash. It avoids duplication of the method name and opens up the possibility of parametrising the tests.

PR #73, opened 2016-08-08
JS test

RibeiroAna:


testing/test.js
```diff
@@ -0,0 +1,6 @@
+QUnit.test( "toArray", function(assert) {
+  function NodeList() {}
+  nodeList = new NodeList();
+  assert.ok(toArray(nodeList) instanceof Array)
+  assert.ok(toArray(null) === null)
+});
```
RibeiroAna: I did just one test just to see how it works

pytest_html/resources/main.js
```diff
@@ -2,6 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+//created to increase testability
+var  DOCUMENT = document;
+
+function setDocument(elem) {
+  DOCUMENT = elem;
+}
```
RibeiroAna: @davehunt I would like to know if you think that is a good practice, because I could not mock document by another way, and I thought that it maybe would be a not that good practice.

testing/test.js
```diff
@@ -0,0 +1,105 @@
...
+  toggle_sort_states(row_numeric);
+  assert.ok(row_numeric.className === 'sortable numeric active desc');
+});
+
+QUnit.test('sort_column_numeric', function(assert){
+  var table = create_elem_table();
+  setDocument(table);
+  row_numeric = find('.numeric', table);
+  sort_column(row_numeric);
+  assert.ok(0 == 0);
```
RibeiroAna: This test is not finished yet, I am waiting for your comments if you think this is a good practice.

.travis.yml
```diff
@@ -19,5 +20,9 @@ matrix:
       env: TOXENV=flake8
 install:
   - pip install tox-travis
+  - npm install -g grunt-cli
+  - npm install grunt
+  - npm install grunt-contrib-qunit
 script:
   - tox
+  - grunt test
```
davehunt: This will run the JavaScript tests for every python/toxenv configuration (16 times). We only need to run it once. We could set up CircleCI to run the JavaScript tests, or we could modify Travis CI so that it only runs these on one configuration. This could possibly be done by setting a `QUNIT` environment variable in the appropriate place and making the appropriate install and script lines dependent on that variable.

package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "pytest-html",
+  "version": "1.10.0",
```
davehunt: Is it necessary to create a package? I'd like to at least avoid maintaining the version number in multiple places, so if we need this please change the version to something like 0.0.1 so it's clear we're not needed to keep it in sync with the python package.

package.json
```diff
@@ -0,0 +1,28 @@
...
+    "url": "git://github.com/pytest-dev/pytest-html.git"
+  },
+  "description": "pytest-html ===========",
+  "bugs": {
+    "url": "https://github.com/pytest-dev/pytest-html/issues"
+  },
+  "homepage": "https://github.com/pytest-dev/pytest-html#readme",
+  "main": "Gruntfile.js",
+  "dependencies": {
+    "grunt-contrib-qunit": "^1.2.0"
```
davehunt: This and the devDependencies appear to be duplicating the installation of dependencies done in the Travis configuration. Is it necessary to have both?

pytest_html/resources/main.js
```diff
@@ -12,18 +18,84 @@ function toArray(iter) {
...
 
 function find_all(selector, elem) {
     if (!elem) {
-        elem = document;
+        elem = DOCUMENT;
     }
     return toArray(elem.querySelectorAll(selector));
 }
 
+function reset_sort_headers() {
```
davehunt: The patch is hard to read because these functions have been moved - could you move them back so the diff shows the lines changed more clearly?

pytest_html/resources/main.js
```diff
@@ -12,18 +18,84 @@ function toArray(iter) {
...
+    //there isn't Unit Tests for this function
```
davehunt: Is there a reason we don't have tests for this? Perhaps there's a test coverage too we could run to see what's missing tests?

pytest_html/resources/main.js
davehunt: I would agree that this is probably not particularly good practice. There are some useful suggestions [here](http://developer.fortnox.se/blog/testing-the-dom-in-javascript/) that basically propose simplifying the JavaScript code to not depend on accessing the DOM directly or by using a QUnit fixture to mock the HTML content.

davehunt: This is also a pretty good resource with some suggestions: https://qunitjs.com/intro/#testing-the-dom-manipulation

.travis.yml
RibeiroAna: I think we've talked about that in Mountain View: we thought that running 16 times would not be a big impact to the runtime, and it would be weird if one pytest fails to js and other no and failing them all would be more intuitive (it is not very intuitive, at the end :P)... But I liked the idea of setting up CircleCI to run JS tests, have you ever done a project with both CI at the same time? How does it works?

package.json
RibeiroAna: I think I've generated this file by some automatic command line interface, I'll check which of those are really necessary.

pytest_html/resources/main.js
RibeiroAna: I think it was just a self remain

.travis.yml
davehunt: It wouldn't be a big impact to runtime, but it's not ideal if we can avoid it. I've asked @nicoddemus to enable this repository with CircleCI so we can move the JS tests there. Until then, you should be able to enable it for your fork and experiment.

pytest_html/resources/main.js
davehunt: Thumbs up and thumbs down? How ambiguous. 😆

.travis.yml
davehunt: CircleCI should be enabled now.

pytest_html/resources/main.js
RibeiroAna: Haha I was going to tumbs up but I did tumps down and I could not undo =P

RibeiroAna: But you can interpreter the tumbs down as: How much work that givens! =P

RibeiroAna:
Hi @davehunt, circleCI is running on mine fork: https://circleci.com/gh/RibeiroAna/pytest-html/3, I think maybe the configuration of this repository is not running for every pull request because here it didn't run on my request..

davehunt:
> it didn't run on my [pull] request

It might be because this pull request was opened before CircleCI was configured. I'll try closing and reopening the pull request to see if that triggers a build.

davehunt:
Hmm.. it didn't trigger a build, but I found this:

> Note: If otherdev/yourproject is explicitly configured to build on CircleCI in its own right, then we will not run builds in the yourorg/yourproject context for PRs from the fork in either case.

From https://circleci.com/docs/fork-pr-builds/

pytest_html/resources/main.js
```diff
@@ -97,7 +97,8 @@ function add_collapse() {
         });
     })
 }
-addEventListener("DOMContentLoaded", function() {
+
+function init () {
```
RibeiroAna: I did that because it helps on testability

RibeiroAna:
I added a coverage tool that doesn't appears on Circle, but on the html report of the JS tests, and its actual coverage is about 95%. I decided to do only tests to the public methods (Aka, methods directly called by HTML) and since the tests to find all and find already existed, I decide to let them there, for sure, we could stress more the tests cases, but I think it is a good start point for that.

screen shot 2016-09-27 at 21 55 47

Gruntfile.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function(grunt) {
+  grunt.initConfig({
+    qunit: {
+      src: 'testing/js_test_report.html'
+    }
+  });
+  grunt.loadNpmTasks('grunt-contrib-qunit');
+  grunt.registerTask('test', 'qunit:src');
+  grunt.registerTask('travis', ['lint', 'test']);
```
davehunt: Does 'travis' refer to Travis CI? Can we get the linter to run in CircleCI too?

testing/js_test_report.html
```diff
@@ -0,0 +1,62 @@
+<!DOCTYPE html>
```
davehunt: I assume we'll need to keep this updated when we change the HTML structure of the report generated?

circle.yml
```diff
@@ -0,0 +1,11 @@
...
+dependencies:
+  override:
+    - npm install -g grunt-cli
+    - npm install grunt
+    - npm install grunt-contrib-qunit
+
+## Customize test commands
+test:
+  override:
+    - grunt test
```
davehunt: Does this generate any reports that we can put into CircleCI's [test results](https://circleci.com/docs/test-metadata/) or [artifacts](https://circleci.com/docs/build-artifacts/)?

testing/js_test_report.html
RibeiroAna: Yes, that is right, if we change the structure, we should change the HTML there

circle.yml
RibeiroAna: it prints something like: 
>> 38 assertions passed (106ms)

Done, without errors.
It is not anything more than what is already there

Gruntfile.js
RibeiroAna: About the coverage, I tried, but I could not.

PR #75, opened 2016-09-10
Add ANSI support in html report

premkarat:
ANSI codes used for color logging is not reatained in html report.
This patch adds that support. Fixes bug # 74

Signed-off-by: Prem Karat pkarat@mvista.com


pytest_html/plugin.py
```diff
@@ -4,6 +4,7 @@
 
 from __future__ import absolute_import
 
+from ansi2html import Ansi2HTMLConverter
```
davehunt: This should be with the other third-party imports, on line 18.

pytest_html/plugin.py
```diff
@@ -225,7 +226,9 @@ def append_log_html(self, report, additional_html):
             for header, content in report.sections:
                 log.append(' {0} '.format(header).center(80, '-'))
                 log.append(html.br())
-                log.append(content)
+                content = Ansi2HTMLConverter(inline=True).convert(content,
+                                                                  full=False)
```
davehunt: This will add inline styles, which won't work if CSP is enforced. Could you try calling `get_styles`(from ansi2html.style import get_styles) to append the generated styles to the existing style.css, or create a separate ansi.css that's linked. If the user specifies `--self-contained` then we could append the styles to the header, which should be better than repeating styles with `inline=True`.

test_pytest_html.py
```diff
@@ -476,3 +476,20 @@ def test_foo(val):
         result, html = run(testdir)
         assert result.ret == 0
         assert_results(html, passed=1)
+
+    def test_ansi_color(self, testdir):
+        pass_content = ["color: #aa0000\">RCOLOR", "color: #00aa00\">GCOLOR",
+                        "color: #aa5500\">YCOLOR"]
+        testdir.makepyfile(r"""
+            import pytest
```
davehunt: No need for this import - it's not used.

test_pytest_html.py
```diff
@@ -476,3 +476,20 @@ def test_foo(val):
...
+            def test_ansi():
+                colors = ['\033[31mRCOLOR\033[0m', '\033[32mGCOLOR\033[0m',
+                          '\033[33mYCOLOR\033[0m']
+                for color in colors:
+                    print(color)
+                pass
```
davehunt: The `pass` is only needed if the method would otherwise be empty.

setup.py
```diff
@@ -10,7 +10,7 @@
       packages=['pytest_html'],
       package_data={'pytest_html': ['resources/*']},
       entry_points={'pytest11': ['html = pytest_html.plugin']},
-      install_requires=['pytest>=2.3'],
+      install_requires=['pytest>=2.3', 'ansi2html>=1.1.1'],
```
davehunt: Please put the requirements on separate lines.

davehunt:
This is a great enhancement, thanks @premkarat - let me know if you have any questions regarding my comments. :)

premkarat:
Thank You Dave, will work on your suggestions and submit an updated patch

premkarat:
Dave,
I did give a try now, and probably its not doing inline anymore. I followed your suggestions as-is. Please correct me if I overlooked something. I didn't have a clue on CSS and through this work I did have some learning experience. Thanks for that.

premkarat:
Hi @davehunt did you get a chance to look into this, Please let me know if further rework/input is required.

davehunt:
This looks good, but I'd like to avoid adding all the CSS from ansi2html as a static file. Instead we can generate it ourselves and append it to our style.css. I just got this to work by adding the following to `_generate_report`:

```python
from ansi2html import style
ansi_css = [
    '\n/******************************',
    ' * ANSI2HTML STYLES',
    ' ******************************/\n']
ansi_css.extend([str(r) for r in style.get_styles()])
self.style_css += '\n'.join(ansi_css)
```

There's also been a few recent changes, so could you rebase your branch? Once this is done I'd be happy to release a v1.11.0 with this awesome enhancement!

premkarat:
Thank You Dev for you inputs and guidance. I didn't get a chance to work on this earlier. Please feel free to let me know if further action is required from my side.

davehunt:
Hey @premkarat, did you rebase and squash/fixup new commits? This makes it difficult to see what's been updated during a review. Fortunately this is a relatively small patch, so it won't be too difficult for me to read through the entire this again.

pytest_html/plugin.py
```diff
@@ -15,6 +15,8 @@
 import bisect
 import hashlib
 
+from ansi2html import Ansi2HTMLConverter
+from ansi2html import style
```
davehunt: Nit: could you put these on a single line:

```python
from ansi2html import Ansi2HTMLConverter, style
```

pytest_html/plugin.py
```diff
@@ -278,16 +282,24 @@ def append_other(self, report):
         self._appendrow('Rerun', report)
 
     def _generate_report(self, session):
+        ansi_css = [
+            '\n/******************************',
+            ' * ANSI2HTML STYLES',
+            ' ******************************/\n']
```
davehunt: Nit: Could you move this down to line 300 so it's immediately before where it's used. I think this makes it a little easier when reading the code.

premkarat:
Sorry Dave for all the confusion. I have created a new pull request here. 
 Add ANSI support in html report #81, https://github.com/pytest-dev/pytest-html/pull/81

PR #80, opened 2016-10-14
Allow displaying multiple images in test report.

### LINKED ISSUE #35: Provide a hook to document arbitrary additional details about each test run.
I'm using `pytest-html` with `pytest-selenium`. My selenium tests are doing an image diff.

I want to include the results of a failed image diff test in my pytest-html report. 

However, if I just use `pytest_html.extras.append(...)` then my output gets mixed up with pytest-selenium's output and makes for a very incoherent report e.g.

screen shot 2016-03-01 at 3 51 00 pm

Full report at: https://s3.amazonaws.com/bokeh-travis/112986254/tests/pytest-report.html

I have hacked around this at the moment by overriding `longrepr` and `extras` completely so that I get something like this:

screen shot 2016-03-01 at 3 42 47 pm

Full report at: https://s3.amazonaws.com/bokeh-travis/112997379/tests/pytest-report.html

What I'd like is a hook to be able to add another row of my own custom test output.

karandesai-96:
This PR attempts to add support for multiple images with a particular method in the report. It aims to fix #35. Results are shown in [this screenshot](https://github.com/pytest-dev/pytest-html/issues/35#issuecomment-253738623).

Adding multiple images is as simple as making a hookwrapper, passing a **list** of base64 encoded strings to `extras`...

``` python
@pytest.mark.hookwrapper
def pytest_runtest_makereport(item, call):
    pytest_html = item.config.pluginmanager.getplugin('html')
    outcome = yield
    report = outcome.get_result()
    extra = getattr(report, 'extra', [])

    if report.when == 'call':
        # report.nodeid was 'lib.linux-x86_64-2.7/tardis/tests/test_util.py::test_quantity_parser_normal'

        if report.nodeid.split('::')[-1] == '::test_quantity_parser_normal':
            # add two images to this method
            with open("/path/to/image1.png", "rb") as image1:
                encoded1 = base64.b64encode(image1.read())
            with open("/path/to/image2.png") as image2:
                encoded2 = base64.b64encode(image2.read())

            extra.append(pytest_html.extras.image(
                [encoded1, encoded2], name=['P-CONTROL', 'PI-CONTROL']
            ))
        report.extra = extra
```

Passing a string to display single image will still work fine.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -140,7 +140,8 @@ class HTMLReport(object):
             # need different hash keys for different images
             if image_index:
                 hash_key = ''.join([self.test_id, str(extra_index),
-                                   str(test_index), str(image_index)]).encode('utf-8')
+                                    str(test_index), str(image_index)]).encode(
+                        'utf-8')
             else:
                 hash_key = ''.join([self.test_id, str(extra_index),
                                    str(test_index)]).encode('utf-8')
@@ -173,11 +174,11 @@ class HTMLReport(object):
                     content = extra.get('content')
                     for image_index, b64_image in enumerate(content):
                         if PY3:
-                            decoded_image = b64decode(b64_image.encode('utf-8'))
+                            decoded = b64decode(b64_image.encode('utf-8'))
                         else:
-                            decoded_image = b64decode(b64_image)
+                            decoded = b64decode(b64_image)
                         single_image_href = src = self.create_asset(
-                            decoded_image, extra_index, test_index, 'png', 'wb',
+                            decoded, extra_index, test_index, 'png', 'wb',
                             image_index=image_index)
                         self.additional_html.append(html.div(
                             html.a(html.img(src=src), href=single_image_href),
@@ -211,11 +212,11 @@ class HTMLReport(object):
                 # for multiple images, extra.get('name') will return a list and
                 # href will be a list, else both of these would be strings
                 if isinstance(href, list):
-                    for single_image_href, name in zip(href, extra.get('name')):
+                    for single_href, name in zip(href, extra.get('name')):
                         self.links_html.append(html.a(
                             name,
                             class_=extra.get('format'),
-                            href=single_image_href,
+                            href=single_href,
                             target='_blank'))
                         self.links_html.append(' ')
                 else:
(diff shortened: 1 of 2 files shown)


karandesai-96:
@davehunt :tada: :tada: There are some hacks, but I am thinking of some generalizations. The code will keep on growing in a single `plugin.py`, so can we split different classes in different files ? It took me a while to figure out the places where I had to do the changes :sweat_smile: 

Also I can extend this to append multiple _hrefs_ if they aren't allowed in current `master`.

karandesai-96:
_In process of making Travis happy again._

davehunt:
It's already possible to add multiple images by calling `extra.append(pytest_html.extras.image(...))` multiple times. The assets will be unique because we already use an index for the extras. For me, just your CSS change of `clear: right;` gives the same results with the following hookwrapper:

```python
import base64

import pytest


@pytest.mark.hookwrapper
def pytest_runtest_makereport(item, call):
    pytest_html = item.config.pluginmanager.getplugin('html')
    outcome = yield
    report = outcome.get_result()
    extra = getattr(report, 'extra', [])
    if report.when == 'call':
        for image in ['image1.png', 'image2.png']:
            with open(image, 'rb') as f:
                encoded = base64.b64encode(f.read()).decode('utf-8')
                extra.append(pytest_html.extras.image(encoded))
        report.extra = extra
```

Everything else seems unnecessary. What's not ideal is that the height of the log is not extended to the full height of the multiple images. Also, if we had more than two images the report would become quite unreadable due to all the whitespace. I think it might actually be worth holding off and presenting the multiple images as a gallery. Perhaps a simple [accordion](http://thecodeplayer.com/walkthrough/make-an-accordian-style-slider-in-css3)?

davehunt:
Actually something [like this](http://tulios.github.io/jquery.silver_track/) would be nice. We should avoid using third party libraries though, and we don't need smooth animations. We should also make sure that if we rely on JavaScript that all images are displayed when it's disabled or restricted. If possible, let's use just CSS.

karandesai-96:
I don't know why this idea didn't strike me, my bad ! My current implementation is an overkill, although these links seem very promising. I'll try to shape up things a bit, meanwhile please keep this PR as is. I'll squash older commits and enter new ones once I get this in position :smile:

davehunt:
Hey @karandesai-96 are you still interested in working on this?

BeyondEvil:
@davehunt This can be closed, right?

davehunt:
Closing in favour of the more recent work in #112

PR #87, opened 2016-11-11
Add support for jpg and svg

bhzunami:
It would be nice if the report support JPEG and SVG images. At the moment it only supports PNG. And if the image is not a png the browser can't display the image properly.

### DIFF
--- a/pytest_html/extras.py
+++ b/pytest_html/extras.py
@@ -8,9 +8,13 @@ FORMAT_JSON = 'json'
 FORMAT_TEXT = 'text'
 FORMAT_URL = 'url'
 
+TYPE_JPG = ('image/jpeg', 'jpg')
+TYPE_PNG = ('image/png', 'png')
+TYPE_SVG = ('image/svg+xml', 'svg')
 
-def extra(content, format, name=None):
-    return {'name': name, 'format': format, 'content': content}
+
+def extra(content, format, name=None, type=None):
+    return {'name': name, 'format': format, 'content': content, 'type': type}
 
 
 def html(content):
@@ -18,7 +22,16 @@ def html(content):
 
 
 def image(content, name='Image'):
-    return extra(content, FORMAT_IMAGE, name)
+    return imagePNG(content=content, name=name)
+
+def imagePNG(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, type=TYPE_PNG)
+
+def imageJPG(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, type=TYPE_JPG)
+
+def imageSVG(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, type=TYPE_SVG)
 
 
 def json(content, name='JSON'):
--- a/pytest_html/extras.py
+++ b/pytest_html/extras.py
@@ -1,6 +1,8 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+import binascii
+import base64
 
 FORMAT_HTML = 'html'
 FORMAT_IMAGE = 'image'
@@ -8,13 +10,10 @@ FORMAT_JSON = 'json'
 FORMAT_TEXT = 'text'
 FORMAT_URL = 'url'
 
-TYPE_JPG = ('image/jpeg', 'jpg')
-TYPE_PNG = ('image/png', 'png')
-TYPE_SVG = ('image/svg+xml', 'svg')
 
-
-def extra(content, format, name=None, type=None):
-    return {'name': name, 'format': format, 'content': content, 'type': type}
+def extra(content, format, name=None, mime_type='image/png', extension='png'):
+    return {'name': name, 'format': format, 'content': content,
+            'mime_type': mime_type, 'extension': extension}
 
 
 def html(content):
@@ -22,19 +21,19 @@ def html(content):
 
 
 def image(content, name='Image'):
-    return imagePNG(content=content, name=name)
+    return png(content=content, name=name)
 
 
-def imagePNG(content, name='Image'):
-    return extra(content, FORMAT_IMAGE, name, type=TYPE_PNG)
+def png(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, mime_type='image/png', extension='png')
 
 
-def imageJPG(content, name='Image'):
-    return extra(content, FORMAT_IMAGE, name, type=TYPE_JPG)
+def jpg(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, mime_type='image/jpeg', extension='jpg')
 
 
-def imageSVG(content, name='Image'):
-    return extra(content, FORMAT_IMAGE, name, type=TYPE_SVG)
+def svg(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, mime_type='image/svg+xml', extension='svg')
 
 
 def json(content, name='JSON'):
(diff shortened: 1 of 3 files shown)


pytest_html/extras.py
```diff
@@ -8,17 +8,33 @@
...
 FORMAT_URL = 'url'
 
+TYPE_JPG = ('image/jpeg', 'jpg')
+TYPE_PNG = ('image/png', 'png')
+TYPE_SVG = ('image/svg+xml', 'svg')
 
-def extra(content, format, name=None):
-    return {'name': name, 'format': format, 'content': content}
+
+def extra(content, format, name=None, type=None):
```
davehunt: I think it would make sense to add `mime_type` and `extension` as optional arguments. Note that `type` is reserved.

pytest_html/extras.py
```diff
@@ -8,17 +8,33 @@
...
+    return {'name': name, 'format': format, 'content': content, 'type': type}
 
 
 def html(content):
     return extra(content, FORMAT_HTML)
 
 
 def image(content, name='Image'):
```
davehunt: We can expose `mime_type` and `extension` as optional arguments, defaulting to the values for PNG.

pytest_html/extras.py
```diff
@@ -8,17 +8,33 @@
...
+def imagePNG(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, type=TYPE_PNG)
+
+
+def imageJPG(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, type=TYPE_JPG)
+
+
+def imageSVG(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name, type=TYPE_SVG)
```
davehunt: I'm undecided on these helper functions. I think allowing users the specify the mime type and extension gives them full flexibility without too much complexity, but I can also see the advantage of them.

davehunt: Let's keep them, but I'd rename them to simply `png` (alias for `image`), `jpg`, and `svg`.

pytest_html/plugin.py
```diff
@@ -159,8 +159,9 @@ def append_extra_html(self, extra, extra_index, test_index):
             href = None
             if extra.get('format') == extras.FORMAT_IMAGE:
                 if self.self_contained:
-                    src = 'data:image/png;base64,{0}'.format(
-                            extra.get('content'))
+                    src = 'data:{0};base64,{1}'.format(
+                        extra.get('type')[0],
```
davehunt: This would become `extras.get('mime_type')`

pytest_html/plugin.py
```diff
@@ -170,7 +171,8 @@ def append_extra_html(self, extra, extra_index, test_index):
                     else:
                         content = b64decode(content)
                     href = src = self.create_asset(
-                        content, extra_index, test_index, 'png', 'wb')
+                        content, extra_index, test_index,
+                        extra.get('type')[1], 'wb')
```
davehunt: This would become `extras.get('extension')`.

davehunt: We can also do this in code where we output json and text files.

pytest_html/extras.py
```diff
@@ -9,16 +9,32 @@
 FORMAT_URL = 'url'
 
 
-def extra(content, format, name=None):
-    return {'name': name, 'format': format, 'content': content}
+def extra(content, format, name=None, mime_type='image/png', extension='png'):
```
davehunt: Here the `mime_type` and `extension` should default to `None`. Then we can set the appropriate values in `image`, `png`, `jpg`, `svg`, `json`, and `text`.

pytest_html/extras.py
davehunt: Add the optional arguments so users can extend this themselves:

```python
def image(content, name='Image', mime_type='image/png', extension='png'):
    return extra(content, FORMAT_IMAGE, name, mime_type, extension)
```

pytest_html/extras.py
```diff
@@ -9,16 +9,32 @@
...
 
 
 def image(content, name='Image'):
-    return extra(content, FORMAT_IMAGE, name)
+    return png(content=content, name=name)
+
+
+def png(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name,
+                 mime_type='image/png', extension='png')
```
davehunt: Use `image` here:

```python
return image(content, name, 'image/png', 'png')
```

pytest_html/extras.py
```diff
@@ -9,16 +9,32 @@
...
+
+
+def jpg(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name,
+                 mime_type='image/jpeg', extension='jpg')
```
davehunt:
```python
return image(content, name, 'image/jpeg', 'jpg')
```

pytest_html/extras.py
```diff
@@ -9,16 +9,32 @@
...
+
+
+def svg(content, name='Image'):
+    return extra(content, FORMAT_IMAGE, name,
+                 mime_type='image/svg+xml', extension='svg')
```
davehunt:
```python
return image(content, name, 'image/svg+xml', 'svg')
```

pytest_html/extras.py
```diff
@@ -9,16 +9,32 @@
...
 
 
 def json(content, name='JSON'):
```
davehunt:
```python
return extra(content, FORMAT_JSON, name, 'text/json', 'json')
```

Then you can use `extra.get('mime_type')` and `extra.get('extension')` in the block starting on line 184 of plugin.py. Same applies to `text`.

testing/test_pytest_html.py
```diff
@@ -323,6 +323,42 @@ def pytest_runtest_makereport(item, call):
         src = 'data:image/png;base64,{0}'.format(content)
         assert '<img src="{0}"/>'.format(src) in html
 
+    def test_extra_jpg_image(self, testdir):
```
davehunt: Could you make `test_extra_image` take parameters instead of creating new tests? I won't block on that, though.

bhzunami:
Thanks for the good inputs. I changed the JSON mime type to application/json. 
I would like to add more tests, specially for the _test_extra_separated_ but I always get a failure if I run the test.

What are the steps to run all tests? Or how I have to setup my environment.
I have a virtuelenv where I installed pytest.
I run the tests with:
`pytest testing`

testing/test_pytest_html.py
```diff
@@ -305,7 +305,12 @@ def pytest_runtest_makereport(item, call):
             content)
         assert link in html
 
-    def test_extra_image(self, testdir):
+    @pytest.mark.parametrize('mime_type, extension',
+                             [('image/png', 'png'),
+                              ('image/png', 'image'),
+                              ('image/jpeg', 'jpg'),
+                              ('image/svg+xml', 'svg')])
+    def test_extra_image(self, testdir, mime_type, extension):
```
davehunt: Thanks for making this parameterised. I wonder though if we could actually have a new test `test_extra` to replace `test_extra_json`, `test_extra_image`, and `test_extra_text`? It would be similar to `test_extra_separated`, which itself could be enhanced to include the new image types. What do you think?

bhzunami: I did not add a `test_extra` yet. I thought the test would be cleaner. Because I have to put two extra `if` in the test.
```
if file_type == 'json:
  content_str = json.dumps(content)
```
If you think this is not a problem, I can merge the test in one.

davehunt: That's fine then, we can always revisit the tests later.

davehunt:
Thanks @bhzunami! I've merged this. Feel free to revisit the tests if you think you can make further improvements. Also, if you fancy adding a note to the README on how to use the new image types, that would be great too!

bhzunami:
Thanks a lot for you help @davehunt. I will add the some Information about the extra to the README.

PR #88, opened 2016-11-11
Added version number and pypi url to report header

denisra:
Created two constants in the __init__ file to hold the VERSION and
PYPI_URL.
This patch fixes issue #86.

### DIFF
--- a/pytest_html/__init__.py
+++ b/pytest_html/__init__.py
@@ -1,0 +1,2 @@
+VERSION = '1.11.1'
+PYPI_URL = 'https://pypi.python.org/pypi/pytest-html'
--- a/pytest_html/__init__.py
+++ b/pytest_html/__init__.py
@@ -1,2 +1,11 @@
-VERSION = '1.11.1'
-PYPI_URL = 'https://pypi.python.org/pypi/pytest-html'
+from pkg_resources import get_distribution, DistributionNotFound
+try:
+    dist = get_distribution(__name__)
+except DistributionNotFound:
+    # package is not installed
+    pass
+else:
+    __version__ = dist.version
+    metadata = dist._get_metadata(dist.PKG_INFO)
+    home_page = [m for m in metadata if m.startswith('Home-page:')]
+    __pypi_url__ = home_page[0].split(':', 1)[1].strip()
--- a/pytest_html/__init__.py
+++ b/pytest_html/__init__.py
@@ -1,11 +1,10 @@
 from pkg_resources import get_distribution, DistributionNotFound
+
+
 try:
-    dist = get_distribution(__name__)
+    __version__ = get_distribution(__name__).version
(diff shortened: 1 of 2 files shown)


hayderimran7:
hi @denisra , this package uses `setuptools_scm` to manage its version. shouldn't it actually get the version number as this https://github.com/pypa/setuptools_scm#setuppy-usage point#3 , which states to get version via `pkg_resources` as following:
```
from pkg_resources import get_distribution, DistributionNotFound
try:
    __version__ = get_distribution(__name__).version
except DistributionNotFound:
   # package is not installed
   pass
```

denisra:
thanks @hayderimran7 and @davehunt for pointing that out. I've sent another commit. Please let me know if I need to make any other changes.

pytest_html/__init__.py
```diff
@@ -0,0 +1,11 @@
+from pkg_resources import get_distribution, DistributionNotFound
+try:
+    dist = get_distribution(__name__)
+except DistributionNotFound:
+    # package is not installed
+    pass
```
davehunt: Note that we'd raise an exception when attempting to access `__version__` or `__pypi_url__`.

pytest_html/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+else:
+    __version__ = dist.version
+    metadata = dist._get_metadata(dist.PKG_INFO)
+    home_page = [m for m in metadata if m.startswith('Home-page:')]
+    __pypi_url__ = home_page[0].split(':', 1)[1].strip()
```
davehunt: This works great, however I just noticed the URL is the GitHub project and not the PyPI homepage. Let's just hard-coded the PyPI homepage URL as https://pypi.python.org/pypi/pytest-html

pytest_html/plugin.py
```diff
@@ -392,7 +393,9 @@ def generate_summary_item(self):
             html.script(raw(main_js)),
             html.p('Report generated on {0} at {1}'.format(
                 generated.strftime('%d-%b-%Y'),
-                generated.strftime('%H:%M:%S'))),
+                generated.strftime('%H:%M:%S')),
+                html.a(' pytest-html', href=__pypi_url__),
```
davehunt: Add ' by ' so it reads 'Report generated on 28-Nov-2016 at 09:17:35 by pytest-html v1.11.1'

pytest_html/__init__.py
denisra: basically just re-raise the Exception?

denisra:
@davehunt here is another update. Please let me know if I missed anything.

davehunt:
Awesome, thanks @denisra!

hayderimran7:
great!! this is to hoping all pytest projects soon start using the awesome setuptools_scm :)

PR #97, opened 2016-12-19
Disable ANSI codes support by default

davehunt:
Due to ansi2html having a less permissive license, remove it as a dependency.


CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Release Notes
 -------------
 
+**1.12.1 (2016-12-19)**
```
nicoddemus: I don't know if pytest-html follows semantic versioning, but if it does, shouldn't this be `1.13.0`? I think it makes sense to make new minor release so people can pin to it instead of pinning to a patch release.

davehunt: Yes, I just noticed this myself. Already fixed. :)

setup.py
```diff
@@ -11,8 +11,7 @@
       package_data={'pytest_html': ['resources/*']},
       entry_points={'pytest11': ['html = pytest_html.plugin']},
       setup_requires=['setuptools_scm'],
-      install_requires=['ansi2html>=1.1.1',
-                        'pytest>=2.3'],
+      install_requires=['pytest>=2.3'],
```
JensTimmerman: isn't this only needed for testing? You could put it in `tests_require` instead of `install_requires`

PR #102, opened 2017-02-23
Escape log sections in HTML report. Fixes #100

### LINKED ISSUE #100: pytest-html isn’t escaping HTML special characters in logs
I’m seeing things like:

    auth.py                    518 DEBUG    Auth User is now <authuser('id:1[default] auth:true')>

in the HTML sources of the generated reports. That’s apparently due to a [missing escape call](/pytest-dev/pytest-html/commit/f84482bc6a28a547602b5f4e07ffc9044ee427b0#commitcomment-20481987).

davehunt:


davehunt:
Merged in https://github.com/pytest-dev/pytest-html/commit/67e8f28bbdf3430b8bace03cb2d664eeb470f14a

Issue #103, opened 2017-02-24
Order/Sorting of tests in the reports


jo-as:
Hello,

is there any way keep to original order of the tests. Currently I can sort for "Result", "Test" or "Duration". Unfortunately I do not have any unsorted order.. (so in the order were the test was executed).

The reason for that: 
we are performing many tests with fixture-parametrization (eg. 20 different tests on 100 test-cases), when several tests fails, it is quite difficult to analyze when we have only the ordered representation, but not the unsorted representation (in the sequence, where the tests were performed).

Thanks,
Best regards,
Josef

davehunt:
I don't believe the test order is guaranteed by pytest at all. Also, if tests are running in parallel it would be difficult to represent the order in which they were run. This did come up before in #20. My suggestion at the time was to add a timestamp column for each test, which could then be sorted. Would this work for you?

jo-as:
Hello,
thanks for your reply. 
The test-order is not guaranteed, thats clear. But it is difficult to follow the results in the html, when the order of the results in the html is not the same as the order of the test-execution.

A column with the execution-timestamp and the possibilty to sort for it would be perfect!

Best regards,
Josef

davehunt:
@jo-as I've fixed this by allowing users to implement their own hooks to modify the results table, and given an example that adds a sortable timestamp. If you have some time, please take a look over #104. Hopefully I can release 1.14.0 soon with this enhancement.

jo-as:
Hello,

I have tried your unreleased version and it works fine. The solution is perfect for our needs now (as we were using 1.9.0, we did not use the other benefits, yet.. especially the "show", "hide" feature).

Thanks for the work, it is a very useful plugin!

Best regards,
Josef

davehunt:
Fixed by adding hooks in bc2e8a35554df68113cd2ae911dc396cd32cfe93 available in v1.14.0.

PR #104, opened 2017-02-24
Add hooks for modifying the test results table

### LINKED ISSUE #34: Docstring as alternative description for test method/class
One more suggest from me. Hope I'm not impudent, well not yet :)

Docstring is a cool feature, and with it we don't have to describe out tests somewhere else istead of test itself. I would suggest to read docstring from class/function and append it to the title here:
https://github.com/davehunt/pytest-html/blob/master/pytest_html/plugin.py#L154
optionally. Optionally, because when someone wants to perform doctest (dunno if there someone which does :) ), the feature will destroy the report.

davehunt:
Potentially fixes #34, #40, and #103.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -128,6 +128,38 @@ conftest.py file:
               extra.append(pytest_html.extras.html('<div>Additional HTML</div>'))
           report.extra = extra
 
+Extra columns
+~~~~~~~~~~~~~
+
+You can add/remove columns by implementing custom hooks for the header and
+rows. The following example :code:`conftest.py` adds a description column with
+the test function docstring, adds a sortable time column, and removes the links
+column:
+
+.. code-block:: python
+
+  from datetime import datetime
+  from py.xml import html
+  import pytest
+
+  @pytest.mark.optionalhook
+  def pytest_html_results_table_header(cells):
+      cells.insert(2, html.th('Description'))
+      cells.insert(0, html.th('Time', class_='sortable time', col='time'))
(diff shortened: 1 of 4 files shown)


pytest_html/plugin.py
```diff
@@ -64,10 +69,11 @@ def pytest_configure(config):
...
     if htmlpath and not hasattr(config, 'slaveinput'):
-        config._html = HTMLReport(htmlpath,
-                                  config.getoption('self_contained_html'),
-                                  config.pluginmanager
-                                  .hasplugin('rerunfailures'))
+        config._html = HTMLReport(
+            htmlpath,
+            config.getoption('self_contained_html'),
+            config.pluginmanager.hasplugin('rerunfailures'),
+            config)
```
The-Compiler: Since you pass the `config` object, you can remove `self_contained` and `has_rerun` (the latter in `TestResult` too).

README.rst
```diff
@@ -148,17 +148,44 @@ column:
...
+
+.. code-block:: python
+
+  import pytest
+
+  @pytest.mark.optionalhook
+  def pytest_html_results_table_html(report, data):
+      if report.passed:
+          del data[:]
+          data.append(html.div('No log output captured.', class_='empty log'))
```
sylvinus: 👍

davehunt:
Merged in bc2e8a35554df68113cd2ae911dc396cd32cfe93 and released in v1.14.0.

Issue #106, opened 2017-03-03
encoding error


micheletest:
I just tried to update to the latest pytest html and I'm getting an encoding error. The line referred to isn't encoded that I see, so perhaps that needs an update. I can try to look more into this later.

```
metadata: {'Python': '2.7.13', 'Driver': 'Remote', 'Capabilities': {'browserName': 'chrome', 'screenResolution': '1440x900'}, 'Server': 'http://selenium-server:4444', 'Base URL': 'http://local.better.com:3000', 'Platform': 'Linux-4.4.23-31.54.amzn1.x86_64-x86_64-with', 'Plugins': {'bdd': '2.18.1', 'variables': '1.5.1', 'selenium': '1.9.1', 'rerunfailures': '2.1.0', 'html': '1.14.1', 'timeout': '1.2.0', 'base-url': '1.3.0', 'metadata': '1.3.0'}, 'Packages': {'py': '1.4.32', 'pytest': '3.0.6', 'pluggy': '0.4.0'}}
rootdir:  inifile: pytest.ini
plugins: variables-1.5.1, timeout-1.2.0, selenium-1.9.1, rerunfailures-2.1.0, metadata-1.3.0, html-1.14.1, bdd-2.18.1, base-url-1.3.0
timeout: 480.0s method: signal
collected 2 items 
 
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/_pytest/main.py", line 98, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/_pytest/main.py", line 133, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 745, in __call__
INTERNALERROR>     return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 339, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 334, in <lambda>
[...]

PR #107, opened 2017-03-07
Test to deal with issue #106

BeyondEvil:
Something I couldn't figure out is how to make sure that the environment this test is run on doesn't have UTF-8 has it's encoding. Then this test will become a false positive.

Also, this is only a test for issue #106 not a fix.

### DIFF
--- a/testing/test_pytest_html.py
+++ b/testing/test_pytest_html.py
@@ -419,6 +419,21 @@ class TestHTML:
             assert link in html
             assert os.path.exists(src)
 
+    def test_extra_text_encoding(self, testdir):
+        testdir.makeconftest("""
+            import pytest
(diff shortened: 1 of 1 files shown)


davehunt:
Thanks @BeyondEvil, I've incorporated this into #108 with a potential fix. Would you mind building from that branch and seeing if this solves the problem for you?

BeyondEvil:
@davehunt Sure thing, Dave! First thing tomorrow ;) If you see me on the IRC channel, feel free to remind me. :)

BeyondEvil:
@davehunt meh, who needs sleep anyway. Fix confirmed!

davehunt:
Thanks @BeyondEvil! I've merged and tagged, so v1.14.2 should be available soon!

Issue #77, opened 2016-09-21
Enhance extras.image to accept a path/URL


lambourn:
Feature enhancement: enhance extras.image to accept a path/URL as an alternative to a base64 encoded PNG.

Implementation: check the content type of the extra content. If it is not base64 then just assume it to be a path/URL. The resulting href/src for the <img /> should be the provided string value as-is (optionally: could check if path starts with file: or http(s))

Background: In my setup, I do not have base64 encoded extra content when the report gets generated. Instead, my test logic already creates an external folder where screenshots are stored during test execution. So essentially I'd like to just link to them using a given file:// or relative URL inside the report, no matter if the report is self-contained or not.

e.g. this

```
extra.append(pytest_html.extras.image('../screenshots/foo_bar.png'))
extra.append(pytest_html.extras.image('file:///screenshots/foo_bar.png'))
extra.append(pytest_html.extras.image('http://foo.bar/screenshots/foo_bar.png'))

```

would result in 3 <img /> tags in the report with their src/href set to the above values.

PR #108, opened 2017-03-08
Always encode content for data URI

davehunt:
Fixes #106

PR #111, opened 2017-05-08
Add ability to add screenshot as file or link (http)

BeyondEvil:
Addresses issue #77 ping @davehunt

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -52,6 +52,8 @@ of a less permissive license, this package is not included as a dependency. If
 you have this package installed, then ANSI codes will be converted to HTML in
 your report.
 
+.. _self-contained:
+
 Creating a self-contained report
 ----------------------------------
 
@@ -64,6 +66,12 @@ convenient when sharing your results. This can be done in the following way:
 .. code-block:: bash
 
    $ pytest --html=report.html --self-contained-html
+
+Images added as files or links are going to be linked as external resources,
+meaning that the standalone report HTML-file may not display these images
+as expected.
+
+The plugin will issue a warning when adding files or links to the standalone report.
 
 Enhancing reports
 -----------------
@@ -94,11 +102,16 @@ Raw HTML    ``extra.html('<div>Additional HTML</div>')``
 Plain text  ``extra.text('Add some simple Text')``
 URL         ``extra.url('http://www.example.com/')``
 Image       ``extra.image(image, mime_type='image/gif', extension='gif')``
-Image       ``extra.image('file:///path/to/file.png')``
+Image       ``extra.image('/path/to/file.png')``
 Image       ``extra.image('http://some_image.png')``
 ==========  ============================================
 
-**Note**: When using ``--self-contained-html``, images added as files or links may not work as expected.
+**Note**: When adding an image from file, the path can be either absolute
+or relative.
+
+**Note**: When using ``--self-contained-html``, images added as files or links
+may not work as expected, see the
+:ref:`Creating a self-contained report <self-contained>` section for more info.
 
 There are also convenient types for several image formats:
 
--- a/README.rst
+++ b/README.rst
@@ -98,6 +98,8 @@ Image       ``extra.image('file:///path/to/file.png')``
 Image       ``extra.image('http://some_image.png')``
 ==========  ============================================
 
+**Note**: When using ``--self-contained-html``, images added as files or links may not work as expected.
+
 There are also convenient types for several image formats:
 
 ============  ====================
--- a/README.rst
+++ b/README.rst
@@ -52,10 +52,8 @@ of a less permissive license, this package is not included as a dependency. If
 you have this package installed, then ANSI codes will be converted to HTML in
 your report.
 
-.. _self-contained:
-
 Creating a self-contained report
-----------------------------------
+--------------------------------
 
 In order to respect the `Content Security Policy (CSP)
 <https://developer.mozilla.org/docs/Web/Security/CSP>`_,
@@ -110,8 +108,8 @@ Image       ``extra.image('http://some_image.png')``
 or relative.
 
 **Note**: When using ``--self-contained-html``, images added as files or links
-may not work as expected, see the
-:ref:`Creating a self-contained report <self-contained>` section for more info.
+may not work as expected, see section `Creating a self-contained report`_ for
+more info.
 
 There are also convenient types for several image formats:
 
--- a/README.rst
+++ b/README.rst
@@ -94,6 +94,8 @@ Raw HTML    ``extra.html('<div>Additional HTML</div>')``
 Plain text  ``extra.text('Add some simple Text')``
 URL         ``extra.url('http://www.example.com/')``
 Image       ``extra.image(image, mime_type='image/gif', extension='gif')``
+Image       ``extra.image('file:///path/to/file.png')``
+Image       ``extra.image('http://some_image.png')``
 ==========  ============================================
 
 There are also convenient types for several image formats:
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -165,7 +165,7 @@ class HTMLReport(object):
                         warnings.warn('Self-contained HTML report '
                                       'includes link to external '
                                       'resource: {}'.format(content))
-                    html_div = html.img(src=content)
+                    html_div = html.a(html.img(src=content), href=content)
                 elif self.self_contained:
                     src = 'data:{0};base64,{1}'.format(
                         extra.get('mime_type'),
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -162,9 +162,9 @@ class HTMLReport(object):
                 if content.startswith(('file', 'http')) or \
                         os.path.isfile(content):
                     if self.self_contained:
-                        warnings.warn('Images added via file or link, '
-                                      'may not work as expected when '
-                                      'using --self-contained-html')
+                        warnings.warn('Self-contained HTML report '
+                                      'includes link to external '
+                                      'resource: {}'.format(content))
                     html_div = html.img(src=content)
                 elif self.self_contained:
                     src = 'data:{0};base64,{1}'.format(
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -13,6 +13,7 @@ import sys
 import time
 import bisect
 import hashlib
+import re
 
 try:
     from ansi2html import Ansi2HTMLConverter, style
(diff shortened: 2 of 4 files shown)


BeyondEvil:
haha, I just can't get a win. Not sure why there's two commits in there. However, I do recognize the additional commit as one of my old PRs (encoding test issue).

pytest_html/plugin.py
```diff
@@ -157,24 +158,24 @@ def create_asset(self, content, extra_index,
         def append_extra_html(self, extra, extra_index, test_index):
             href = None
             if extra.get('format') == extras.FORMAT_IMAGE:
-                if self.self_contained:
+                content = extra.get('content')
+                if re.match('^file|^http', content):
+                    html_div = html.img(src=content)
+                elif self.self_contained:
```
davehunt: hmm... so this means if the `--self-contained-html` flag is passed, these images will not be displayed in the report at all. Whilst this is possibly acceptable (I don't think we should be downloading/reading these files) it could be confusing to users. Perhaps we could add a warning to the documentation that standalone reports may still include assets referenced by URL? We could also log a warning for each URL based asset when we're in standalone mode.

testing/test_pytest_html.py
```diff
@@ -418,6 +418,39 @@ def test_fail():
             assert link in html
             assert os.path.exists(src)
 
+    @pytest.mark.parametrize('type', ["https://", "file://"])
+    def test_extra_image_non_b64(self, testdir, type):
+        content = type
```
davehunt: `type` is reserved, so we should pick another name for this parameter.

pytest_html/plugin.py
```diff
@@ -157,24 +158,24 @@ def create_asset(self, content, extra_index,
         def append_extra_html(self, extra, extra_index, test_index):
             href = None
             if extra.get('format') == extras.FORMAT_IMAGE:
-                if self.self_contained:
+                content = extra.get('content')
+                if re.match('^file|^http', content):
```
davehunt: Note that this won't cover absolute or relative paths.

BeyondEvil: Not sure what you mean... What won't cover absolute or relative paths? What other types are there? While testing this I used an image with an absolute path.

Do you want to cover relative paths as well?

pytest_html/plugin.py
BeyondEvil: I'll add a warning to the docs and a logging warning (using pytest warning?).

pytest_html/plugin.py
davehunt: I mean if I passed `/absolute/path/to/file`, `C:\absolute\path\to\file`, or `../relative/path/to/file` then this would not work. Perhaps simply using `os.path` to check if the content is a valid path, or we could leave this as a future enhancement.

davehunt: Rather than importing `re` you could just use [`startswith`](https://docs.python.org/3.6/library/stdtypes.html#str.startswith).

BeyondEvil: Got it, I can give it a try, and if it gets to complex we'll leave it for the future.

BeyondEvil: Haha, I started with that but thought that re was better looking :P 
However, with the abs/rel path change, I'll revert back to startswith 👍

pytest_html/plugin.py
BeyondEvil: @davehunt I need some guidance on how to add the pytest warning. Too many options. Do I get the terminal reporter via getplugin()? Do I use one of the hooks? etc.

davehunt: I would use:

```python
import warnings
warnings.warn('blah blah blah')
```

nicoddemus: Just FYI, pytest-3.1 (to be released this week) automatically shows warnings such as these in a summary at the end of the test suite, so this is the best suggestion at the moment. 👍

BeyondEvil: Wording OK with you, @davehunt ?

```                content = extra.get('content')
                if content.startswith(('file', 'http')) or os.path.isfile(content):
                    if self.self_contained:
                        warnings.warn('Images added via file or link, may not '
                                      'work as expected when using '
                                      '--self-contained-html')
                    html_div = html.img(src=content)
                elif self.self_contained:```

Comment (BeyondEvil) [inline pytest_html/plugin.py]: "so this means if the --self-contained-html flag is passed, these images will not be displayed in the report at all. "

Hmm... they should be, whatever html.img() we create gets added after the if/elifs.

File: README.rst [hunk 4]
```diff
@@ -94,8 +94,12 @@ Raw HTML    ``extra.html('<div>Additional HTML</div>')``
 Plain text  ``extra.text('Add some simple Text')``
 URL         ``extra.url('http://www.example.com/')``
 Image       ``extra.image(image, mime_type='image/gif', extension='gif')``
+Image       ``extra.image('file:///path/to/file.png')``
+Image       ``extra.image('http://some_image.png')``
 ==========  ============================================
 
+**Note**: When using ``--self-contained-html``, images added as files or links may not work as expected.
```
davehunt: I think we expand this a little more. These images are going to be linked as external resources even when `--self-contained-html` is specified. The expectation is that this command line option means there's a standalone HTML file that contains all resources.

davehunt: I also think we should move this to the 'Creating a self-contained report' section, and reference/link to that section here.

pytest_html/plugin.py
```diff
@@ -157,24 +158,29 @@ def create_asset(self, content, extra_index,
...
             href = None
             if extra.get('format') == extras.FORMAT_IMAGE:
-                if self.self_contained:
+                content = extra.get('content')
+                if content.startswith(('file', 'http')) or \
+                        os.path.isfile(content):
+                    if self.self_contained:
+                        warnings.warn('Images added via file or link, '
+                                      'may not work as expected when '
+                                      'using --self-contained-html')
```
davehunt: I would prefer something like "Self-contained HTML report includes link to external resource: /foo/bar"

testing/test_pytest_html.py
```diff
@@ -418,6 +418,39 @@ def test_fail():
             assert link in html
             assert os.path.exists(src)
 
+    @pytest.mark.parametrize('src_type', ["https://", "file://"])
+    def test_extra_image_non_b64(self, testdir, src_type):
```
davehunt: For coverage, we should also include a test that creates a file in the testdir and references it as an image path.

testing/test_pytest_html.py
```diff
@@ -418,6 +418,39 @@ def test_fail():
...
+                if report.when == 'call':
+                    from pytest_html import extras
+                    report.extra = [extras.image('{0}')]
+        """.format(content))
+        testdir.makepyfile('def test_pass(): pass')
+        result, html = run(testdir, 'report.html')
+        assert result.ret == 0
+        assert '<img src="{0}"/>'.format(content) in html
+
+    def test_extra_text_encoding(self, testdir):
```
davehunt: I think this test has sneaked into your patch somehow. :)

BeyondEvil: I wrote this way back when you were having encoding issues. I can remove it if you don't want it.

https://github.com/pytest-dev/pytest-html/pull/107

BeyondEvil: ping @davehunt

BeyondEvil: I've started looking at the image/slider PR @davehunt, I've got most of your requests in place, I just need to fine tune the CSS. 

However, I'd prefer to get this PR out of the way first, since they hit more or less exactly the same code.

davehunt: I landed a change to the tests that covered this in https://github.com/pytest-dev/pytest-html/pull/108/files so this test isn't required. You can leave it here and I'll remove it before merging, if you prefer.

README.rst
```diff
@@ -94,8 +102,17 @@ Raw HTML    ``extra.html('<div>Additional HTML</div>')``
...
+Image       ``extra.image('/path/to/file.png')``
+Image       ``extra.image('http://some_image.png')``
 ==========  ============================================
 
+**Note**: When adding an image from file, the path can be either absolute
+or relative.
+
+**Note**: When using ``--self-contained-html``, images added as files or links
+may not work as expected, see the
+:ref:`Creating a self-contained report <self-contained>` section for more info.
```
davehunt: You don't need to create a reference as all titles can already be referenced. This would work as:

```rst
**Note**: When using ``--self-contained-html``, images added as files or links
may not work as expected, see section `Creating a self-contained report`_ for
more info.
```

README.rst
```diff
@@ -52,6 +52,8 @@ of a less permissive license, this package is not included as a dependency. If
 you have this package installed, then ANSI codes will be converted to HTML in
 your report.
 
+.. _self-contained:
```
davehunt: This isn't needed, as the title can already be referenced.

README.rst
```diff
@@ -52,6 +52,8 @@ of a less permissive license, this package is not included as a dependency. If
...
+
 Creating a self-contained report
 ----------------------------------
```
davehunt: Mind trimming the underline here so it doesn't exceed the title length? 👼

README.rst
```diff
@@ -128,6 +145,14 @@ conftest.py file:
...
           report.extra = extra
 
+You can also specify the :code:`name` argument for all types other than :code:`html` which will change the title of the
+created hyper link:
+
+.. code-block:: python
+
+    extra.append(pytest_html.extras.text('some string', name='Different title'))
+
+
```
davehunt: Looks like this came in through a merge/rebase? I can fix this when we come to merge if you like.

pytest_html/plugin.py
```diff
@@ -157,24 +158,29 @@ def create_asset(self, content, extra_index,
...
             if extra.get('format') == extras.FORMAT_IMAGE:
-                if self.self_contained:
+                content = extra.get('content')
+                if content.startswith(('file', 'http')) or \
+                        os.path.isfile(content):
+                    if self.self_contained:
+                        warnings.warn('Self-contained HTML report '
+                                      'includes link to external '
+                                      'resource: {}'.format(content))
+                    html_div = html.img(src=content)
```
davehunt: There's no link here to the image. We should set `href` to the value of `content` and include it in the `html_div`.

testing/test_pytest_html.py
```diff
@@ -418,6 +418,41 @@ def test_fail():
...
+                if report.when == 'call':
+                    from pytest_html import extras
+                    report.extra = [extras.image('{0}')]
+        """.format(content))
+        testdir.makepyfile('def test_pass(): pass')
+        if src_type == "image.png":
+            testdir.makefile('.png', image='pretty picture')
+        result, html = run(testdir, 'report.html')
+        assert result.ret == 0
+        assert '<img src="{0}"/>'.format(content) in html
```
davehunt: Let's also assert the href attribute.

testing/test_pytest_html.py
BeyondEvil: Removed.

README.rst
BeyondEvil: Yes, please. I have no idea if it belongs or not.

pytest_html/plugin.py
BeyondEvil: I'll fix that here, but that is taken care of in the image gallery PR. 👍

BeyondEvil:
Ping @davehunt New PR, hope it is to your satisfaction. :)

davehunt:
This looks great. I'll do a bit of testing locally with it before merging.

davehunt:
Merged in https://github.com/pytest-dev/pytest-html/commit/f19a86d1e3bc48e46336432a0b2dc6b6dc9cfdb0

Issue #110, opened 2017-05-05
Add links to images in self contained report


davehunt:
When `--self-contained-html` is used we should provide links to any images in the HTML report. We should avoid duplicating the large data URI values, so it makes sense to use JavaScript to enable this functionality.

The links in the report column could be hidden by default and displayed via JavaScript, and the href destination could be taken from the associated img src. Similarly, the img could be an inactive link, enabled via JavaScript if enabled.

PhilippSelenium:
I should have read the comment before I commited my pull request :(
In this case I can't fix it. i don't know any javascript.

PR #112, opened 2017-05-10
Add image galler/slider for screenshots etc

BeyondEvil:
I would suggest merging this PR https://github.com/pytest-dev/pytest-html/pull/111 first. Since this PR really hits the same code.

Hopefully this will get the discussion started @davehunt :)

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -105,11 +105,50 @@ class HTMLReport(object):
             self.logfile = logfile
             self.config = config
             self.row_table = self.row_extra = None
+            self.images = []
 
             test_index = hasattr(report, 'rerun') and report.rerun + 1 or 0
 
             for extra_index, extra in enumerate(getattr(report, 'extra', [])):
                 self.append_extra_html(extra, extra_index, test_index)
+
+            num_of_images = len(self.images)
+            if num_of_images == 1:  # Only one image, no need for carousel
+                src, name = self.images.pop()
+                self.additional_html.append(
+                    html.div(html.a(html.img(src=src, alt=name),
+                                    href=src, target='_blank'),
+                             class_='image'))
+
+            elif num_of_images > 1:  # More than one image, build carousel
+                image_html = []
+                for image in range(num_of_images):
+                    carousel_html = []
+                    src, name = self.images[image]
+
+                    # Add (current) image
+                    carousel_html.append(html.a(html.img(
+                        src=src, alt=name), href=src, target='_blank'))
+
+                    # Add previous image
+                    prev_image = (image - 1) % num_of_images
+                    carousel_html.append(
+                        html.a('<', class_='button prev',
+                               href='#{}'.format(self.images[prev_image][0])))
+
+                    # Add next image
+                    next_image = (image + 1) % num_of_images
+                    carousel_html.append(
+                        html.a('>', class_='button next',
+                               href='#{}'.format(self.images[next_image][0])))
+                    image_html.append(
+                        html.div(carousel_html, id='{}'.format(src)))
+                self.additional_html.append(
+                    html.div(html.div(image_html, class_='carousel_inner'),
+                             class_='carousel'))
+
+            else:
+                pass  # No images, append nothing
 
             self.append_log_html(report, self.additional_html)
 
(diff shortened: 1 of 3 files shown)


pytest_html/plugin.py
```diff
@@ -172,9 +184,8 @@ def append_extra_html(self, extra, extra_index, test_index):
                     href = src = self.create_asset(
                         content, extra_index, test_index,
                         extra.get('extension'), 'wb')
-                    self.additional_html.append(html.div(
-                        html.a(html.img(src=src), href=href),
-                        class_='image'))
+                self.images_html.append(html.a(html.img(
+                    src=src), href=href, target='_blank'))
```
davehunt: In the case that the `src` is a base64 encoded string, we're duplicating it for the `href`. This will cause considerable increase in the file size if there's more than a few screenshots. This is actually related to https://github.com/pytest-dev/pytest-html/issues/110, where I suggest resolving this using JavaScript.

BeyondEvil: Keep as is, and revisit in #110 ?

BeyondEvil:
I do _not_ understand how the remote version of this branch can have the "name" PR. I really don't. Because I haven't pushed it. And if it has it, why then do I need to push it again for it to not complain when I'm trying to push real changes. It. Does. Not. Make. Sense.

To avoid this crap, do I push with `-f` ? Because I don't really care about the state of the remote version, I care about the updated changes I have locally.

BeyondEvil:
Ping @davehunt

davehunt:
Also, I just wanted to note that I haven't reviewed the code as I think we should get the style and functionality working first.

BeyondEvil:
That's so weird as I specifically tested that scenario. I also mixed failures with 1 image, with failures with multiple images, and passes. And I couldn't see any issues. Maybe you can screenshot and/or be more specific to what issues you're seeing?

I will look at the jumping around issue (didn't notice that myself).

The URL changing is unfortunately something we have to live with unless we want to involve JS. At least to my knowledge (and the frontenders where I work).

I'll see how close to your mock I can get. :)

BeyondEvil:
Ping @davehunt I'm finally back to normalcy. Well, somewhat. Please have a look at my previous comments.

I'm going to spend some time on this now, but need to get my head back in the game. :)

davehunt:
@BeyondEvil are there any particular comments you'd like me to review?

PhilippSelenium:
Is there any update on this merge request? I would really like to have the ability to click the thumbnail instead of right click --> open image location.

BeyondEvil:
@PhilippSelenium Agreed, that should be an easy fix. I might do a separate PR just to fix that, unless you're willing to do it?

BeyondEvil:
Hi @RibeiroAna !

Would you mind talking a look at this PR? I have to admit it's a bit over my head. :(

The way a report looks today with multiple screenshots is no good. The stacktrace section is all squished together.

https://dporganizer-test-results.s3.amazonaws.com/e2e/2019-01-22_16-56-31_e2e_test_report_24396.html

RibeiroAna:
I'm taking a look @BeyondEvil

RibeiroAna:
@BeyondEvil I think I figured out the problem: when there are multiple test cases, the id for the multiple generated `<div>` were the same, so the Carousel couldn't work.  My suggestion is adding the `test_id` to the id of the div, so it can be a single id per test (of course, you will need to change the "previous" and "next" links as well), I tested locally and this change seemed to work! 

I reinforce Dave's request to style as he mocked because in the way you styled, depending on the image, the "next" and "previous" links are basically invisible, when I tested with some local images with gray background, I couldn't see them. 

Last but not least, I think it would be nice to have a test with the rerun plugin.

pytest_html/plugin.py
```diff
@@ -105,12 +105,51 @@ def __init__(self, outcome, report, logfile, config):
...
+
+                    # Add (current) image
+                    carousel_html.append(html.a(html.img(
+                        src=src, alt=name), href=src, target='_blank'))
+
+                    # Add previous image
+                    prev_image = (image - 1) % num_of_images
+                    carousel_html.append(
+                        html.a('<', class_='button prev',
+                               href='#{}'.format(self.images[prev_image][0])))
```
RibeiroAna:
```suggestion
                               href='#{}'.format(self.images[prev_image][0] + self.test_id)))
```

pytest_html/plugin.py
```diff
@@ -105,12 +105,51 @@ def __init__(self, outcome, report, logfile, config):
...
             self.row_table = self.row_extra = None
+            self.images = []
 
             test_index = hasattr(report, 'rerun') and report.rerun + 1 or 0
 
             for extra_index, extra in enumerate(getattr(report, 'extra', [])):
                 self.append_extra_html(extra, extra_index, test_index)
 
+            num_of_images = len(self.images)
+            if num_of_images == 1:  # Only one image, no need for carousel
```
PhilippSelenium: Have you considered generalizing it using one for loop instead of the if elif else stuff.

pytest_html/plugin.py
```diff
@@ -105,12 +105,51 @@ def __init__(self, outcome, report, logfile, config):
...
+                src, name = self.images.pop()
+                self.additional_html.append(
+                    html.div(html.a(html.img(src=src, alt=name),
+                                    href=src, target='_blank'),
+                             class_='image'))
+
+            elif num_of_images > 1:  # More than one image, build carousel
+                image_html = []
+                for image in range(num_of_images):
```
PhilippSelenium:
```
for i in range(iterable):
    iterable[i]
```
Doesn't look very pythonic
Have a look here: https://stackoverflow.com/questions/1011938/python-previous-and-next-values-inside-a-loop

BeyondEvil:
Thank you so much for your input @PhilippSelenium and @RibeiroAna !

Much appreciated! If your fingers are itching, I'm more than happy to relinquish this PR to your capable hands. If not, I should be able to work some on it next month.

RibeiroAna:
@BeyondEvil I'm not interested on finishing it. If depends on me, you can go on.

ssbarnea:
@BeyondEvil Can you fix conflicts or close it if no longer needed? Thanks.

BeyondEvil:
This is no part of v4.

PR #115, opened 2017-05-15
Doc addition

liiight:
Added example of using `name` param


liiight:
Btw, I'd still like a way to create assets to be used in custom column. Maybe now that I forked the repo I'll take a stab at it

Issue #114, opened 2017-05-11
INTERNALERROR> UnicodeEncodeError: 'ascii' codec can't encode characters in position 496-497: ordinal not in range(128)


tomga:
On some data exception from subject occurs. Part of backtrace is following:
```
INTERNALERROR>   File "/usr/local/pyenv/local/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 339, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
INTERNALERROR>   File "/usr/local/pyenv/local/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 334, in <lambda>
INTERNALERROR>     _MultiCall(methods, kwargs, hook.spec_opts).execute()
INTERNALERROR>   File "/usr/local/pyenv/local/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 614, in execute
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/usr/local/pyenv/local/lib/python2.7/site-packages/pytest_html/plugin.py", line 462, in pytest_runtest_logreport
INTERNALERROR>     self.append_failed(report)
INTERNALERROR>   File "/usr/local/pyenv/local/lib/python2.7/site-packages/pytest_html/plugin.py", line 274, in append_failed
INTERNALERROR>     self._appendrow('Failed', report)
INTERNALERROR>   File "/usr/local/pyenv/local/lib/python2.7/site-packages/pytest_html/plugin.py", line 246, in _appendrow
INTERNALERROR>     result = self.TestResult(outcome, report, self.logfile, self.config)
INTERNALERROR>   File "/usr/local/pyenv/local/lib/python2.7/site-packages/pytest_html/plugin.py", line 113, in __init__
INTERNALERROR>     self.append_log_html(report, self.additional_html)
INTERNALERROR>   File "/usr/local/pyenv/local/lib/python2.7/site-packages/pytest_html/plugin.py", line 216, in append_log_html
INTERNALERROR>     for line in str(report.longrepr).splitlines():
INTERNALERROR> UnicodeEncodeError: 'ascii' codec can't encode characters in position 496-497: ordinal not in range(128)
```

I've managed to fix (or maybe just workaround) this issue locally with following diff:
```
diff --git a/pytest_html/plugin.py b/pytest_html/plugin.py
index afc6275..95814d7 100644
[...]

PR #117, opened 2017-05-24
Add a contributing section to the README

nicoddemus:
As discussed in #114

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -202,6 +202,18 @@ Screenshots
    :target: https://cloud.githubusercontent.com/assets/122800/11951695/f371b926-a88a-11e5-91c2-499166776bd3.png
    :alt: Enhanced HTML report
 
+Contributing
+------------
+
+Fork the repository and submit PRs with bug fixes and enhancements,  contributions are very welcome.
+
+Tests can be run locally with `tox`_, for example to execute tests for Python 2.7 and 3.6 execute::
+
+    tox -e py27,py36
+
+
+.. _`tox`: https://tox.readthedocs.org/en/latest/
+
 Resources
 ---------
 


README.rst
```diff
@@ -202,6 +202,18 @@ Screenshots
...
    :alt: Enhanced HTML report
 
+Contributing
+------------
+
+Fork the repository and submit PRs with bug fixes and enhancements,  contributions are very welcome.
+
+Tests can be run locally with `tox`_, for example to execute tests for Python 2.7 and 3.6 execute::
+
+    tox -e py27,py36
```
davehunt: The tox environment names include the two latest major stable pytest versions (needing an update since 3.1 was released). For example py36-pytest30. Perhaps change the example, or mention the `-l` option to list the available environments?

davehunt: Hmm... it appears that py36 does work.. not sure if this is a bug or a feature.

nicoddemus: > The tox environment names include the two latest major stable pytest versions (needing an update since 3.1 was released).

I noticed this, but also noticed that `-e py27,py35` worked and used the latest pytest version, so just went with it because it seemed enough for doing a local test before submitting a PR. But if you prefer we can bless a test environment for local testing, say `-e py27-pytest31,py36-pytest31`.

FWIW we usually recommend `-e py27,py35,linting` on pytest.

>  it appears that py36 does work.. not sure if this is a bug or a feature.

To be honest I only tested `py35` and it worked, and as an afterthought I changed to `py36`. When you say it does not work, do you mean It does not run at all or it fails some tests?

davehunt: It does work, I was just expecting tox to complain that it's not a valid environment. I'm fine with users running one or both of these environments before submitting a pull, so let's leave this as it is. Thanks for the patch!

nicoddemus: Great, thanks! 👍

PR #120, opened 2017-06-09
Use latest dependencies when running tests

davehunt:
Only run tests against the latest versions of all dependencies.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,16 +4,14 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py{27,36,py,py3}-pytest{29,30}{,-ansi2html}, flake8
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -4,16 +4,14 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py{27,36,py,py3}-pytest{29,30}{,-ansi2html}, flake8
+envlist = py{27,36,py,py3}{,-ansi2html}, flake8
 
 [testenv]
-commands = py.test -v -r a {posargs}
+commands = pytest -v -r a {posargs}
```
nicoddemus: 😆

davehunt: #dropthedot 😄

Issue #122, opened 2017-06-09
New 1.15.0 version gives AttributeError: 'TestReport' object has no attribute 'longreprtext'


silvana-i:
With latest version of pytest-html 1.15.0 we are seeing this error:
`INTERNALERROR> AttributeError: 'TestReport' object has no attribute 'longreprtext'`
Same issue is not visible with the previous 1.14.2.

We use it in its most straight forward way. Calling it from py.test with `pytest --html=report.html`.

nicoddemus:
1.15.0 now pytest 3, unfortunately this was not specified in `setup.py` (should be fixed by #123).

davehunt:
I've merged #123 and tagged for a release of 1.15.1. Thanks @silvana-i and @nicoddemus!

nicoddemus:
🎆

PR #123, opened 2017-06-09
Require pytest >= 3.0 to install

nicoddemus:
Since #121 pytest < 3 is no longer supported

Related to #122


nicoddemus:
After this I we should cut out a new release, otherwise others might experience #122

davehunt:
Thanks @nicoddemus!

nicoddemus:
👍

PR #132, opened 2017-09-11
Fix render metadata list value

rasmuspeders1:
I wanted to show a lists of file names in the metadata view but they were concatenated together in one long string. I think it would be more useful if list like types are rendered as strings with comma seperators.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -441,7 +441,7 @@ class HTMLReport(object):
             if isinstance(value, basestring) and value.startswith('http'):
                 value = html.a(value, href=value, target='_blank')
             elif isinstance(value, (list, tuple, set)):
-                value = ', '.join(value)
+                value = ', '.join((str(i) for i in value))
             rows.append(html.tr(html.td(key), html.td(value)))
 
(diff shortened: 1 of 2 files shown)


davehunt:
@rasmuspeders1 thanks for the pull request! Would you mind also adding a test for this enhancement?

rasmuspeders1:
Of course. test added now.

pytest_html/plugin.py
```diff
@@ -440,6 +440,8 @@ def _generate_environment(self, config):
             value = metadata[key]
             if isinstance(value, basestring) and value.startswith('http'):
                 value = html.a(value, href=value, target='_blank')
+            elif isinstance(value, (list, tuple, set)):
+                value = ', '.join(value)
```
davehunt: This fails with `TypeError: sequence item 0: expected str instance, int found` if `value` is a collection of integers.

rasmuspeders1: Fixed by converting all items in collection to strings before join.

davehunt:
Merged and pushed tag. v1.16.0 should be available shortly.

Issue #124, opened 2017-07-11
path too long on windows when saving report with screenshots


ngavrish:
I have following code to generate screenshots in reports
```
        if xfail or report.failed:
            **extra.append(pytest_html.extras.image(pytest.driver.get_screenshot_as_base64(), ''))**
            extra.append(pytest_html.extras.html('<div><h2 style="color: red; ">{}</h2></div>'.format(test_name)))
        elif report.passed:
            extra.append(pytest_html.extras.html(
                '<div><h2 style="color: black; text-decoration: underline">{}</h2></div>'.format(test_name)))
        pytest_html.extras.html('<div style="color: black; margin: 10px;">'.format(test_name))
        for step in test_steps:
            extra.append(pytest_html.extras.html('<p style="margin: 5px;">{}</p>'.format(step)))
        pytest_html.extras.html('</div>')
        report.extra = extra
```

And you have line in plugin.py where you get file content and validate if its path. And in my case file content contains binary representation of image. When windows is trying to validate this long path, it fails into INTERNAL ERROR.

INTERNALERROR>   File "c:\python36\lib\site-packages\pytest_html\plugin.py", line 112, in __init__
INTERNALERROR>     self.append_extra_html(extra, extra_index, test_index)
INTERNALERROR>   File "c:\python36\lib\site-packages\pytest_html\plugin.py", line 164, in append_extra_html
INTERNALERROR>     os.path.isfile(content):
INTERNALERROR>   File "c:\python36\lib\genericpath.py", line 30, in isfile
INTERNALERROR>     st = os.stat(path)
INTERNALERROR> ValueError: stat: path too long for Windows

davehunt:
Thanks for the bug report. It sounds like we need to catch this exception.

camillake:
@davehunt This issue could be resolved, please refer to https://github.com/pytest-dev/pytest/issues/291. More about windows path: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx

davehunt:
pytest-html [1.16.1](https://pypi.python.org/pypi/pytest-html/1.16.1) has been released with a fix.

PR #136, opened 2017-11-08
Fix for including a screenshot bug on Windows (#124)

pinkie1378:


nicoddemus:
LGTM, the failure seems unrelated though.

pinkie1378:
The Node.js tests didn't run because of a [problem during PhantomJS installation](https://travis-ci.org/pytest-dev/pytest-html/jobs/299359717). This is unrelated to the proposed change. Please advise.

nicoddemus:
@pinkie1378 I don't there's anything to be done in this PR, let's wait for @davehunt's input (he will be back next week it seems).

pytest_html/plugin.py
```diff
@@ -159,8 +160,14 @@ def append_extra_html(self, extra, extra_index, test_index):
...
-                if content.startswith(('file', 'http')) or \
-                        os.path.isfile(content):
+                try:
+                    is_file = (content.startswith(('file', 'http')) or
+                               isfile(content))
+                except ValueError:
+                    # On Windows, os.path.isfile throws this exception when
+                    # passed a b64 encoded image.
+                    is_file = False
+                if is_file:
```
davehunt: I'm okay with this patch, but the `is_file` variable name is misleading. The variable determines if the content is a URI or a path, and not specifically if it's a file. Perhaps we could use a better variable name, or change the logic to determine if the content is a base64 encoded string only falling back to creating a link when it's not.

davehunt:
> The Node.js tests didn't run because of a problem during PhantomJS installation. This is unrelated to the proposed change. Please advise.

If anyone has time to investigate that issue it would be appreciated. I had hoped to be home by now but my travel has been disrupted so I may not get time to look for a few days at best. I wouldn't hold up this patch though, if the failure is clearly unrelated.

davehunt:
I've fixed the Travis CI failure in 764ef3cbbe1dfbe22587ada4856fd67c78603558 so once the requests changes have been made I'm happy to merge and release a new version. Thanks @pinkie1378 and @nicoddemus for your contributions.

davehunt:
pytest-html [1.16.1](https://pypi.python.org/pypi/pytest-html/1.16.1) has been released with this fix.

PR #150, opened 2018-03-03
Added option to append one or more user-provided style sheets to report's styling.

i-am-david-fernandez:
Caveat: this is my first pull request, so opening apologies if I've not done things correctly. Constructive criticism on anything is welcome.

I'm not sure if I've approached the error reporting (failure to read user-supplied CSS file) appropriately, so feel free to alter that.

The approach I've taken allows multiple user-supplied CSS files to be added in the order specified. Perhaps this isn't necessary, but the difference is near trivial, and it provides quite a good increase in flexibility for essentially no cost. It may be useful if a team wants to apply something like business and team/project/client styling.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -55,6 +55,9 @@ def pytest_addoption(parser):
                     'that the report may not render or function where CSP '
                     'restrictions are in place (see '
                     'https://developer.mozilla.org/docs/Web/Security/CSP)')
+    group.addoption('--append-css', action='append', dest='appendcss',
+                    metavar='path', default=None,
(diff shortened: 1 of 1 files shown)


davehunt:
In addition to my review comments, could you look into getting this passing the flake8 checks? You can run them locally using `tox -e flake8`. Also, if you could add tests and documentation for the new feature that would be great! If not, I can take care of these.

davehunt:
Thanks again @i-am-david-fernandez for kicking this off. I've addressed my comments, and will now close this in favour of #156, which includes your valued contribution!

PR #151, opened 2018-03-22
adds hook to edit metadata before writing to Environment section

j19sch:
Running tests on our CI environment creates a lot of metadata and thus a huge Environment table in the report. So I'd like to be able to filter which metadata ends up in the Environment table, so I added a hook to edit the metadata before it's processed by the plugin.

This is my first PR to an open source repo. If I'm not submitting this the proper way, please let me know and I'll fix it. Thanks!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -85,6 +85,17 @@ via the :code:`pytest_configure` hook:
   def pytest_configure(config):
       config._metadata['foo'] = 'bar'
 
+You can make changes to the metadata before it's written to the *Environment* section via the :code:`pytest_html_environment_section_metadata` hook. The following example :code:`conftest.py` removes metadata entries of which the key contains "password":
+
+.. code-block:: python
+
+  @pytest.mark.optionalhook
+  def pytest_html_environment_section_metadata(metadata):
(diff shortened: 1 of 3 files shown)


j19sch:
Will make a similar PR for pytest-metadata.

I can imagine someone wanting to have the metadata during the test run, but not have it in the report. Would you accept a PR that implements a similar hook for the Environment section as you suggested for the Summary section in PR https://github.com/pytest-dev/pytest-html/pull/152?

In addition, I could implement the three existing Summary hooks for the Environment section, so both would basically have the same hooks. Not sure how much real value that would have, though.

j19sch:
PR for pytest-metadata: https://github.com/pytest-dev/pytest-metadata/pull/10

j19sch:
Closing this one, since
- editing the actual metadata can be done via the PR I submitted in pytest-html.
- I don't see enough value in adding an optional prefix and suffix to the Environment section.
- adding a hook to edit the Environment section only has a purpose in case someone reports that editing the metadata itself is not an option for them.

davehunt:
Thanks, we can always revisit additional hooks if users find a need for them.

PR #152, opened 2018-03-25
Additional content in summary section

### LINKED ISSUE #109: How to add additional content to Summary section?
I need to add a Score value to the summary section in html report. There is a mention about how to add additional content to Environment section but not for summary. Please guide me on this.

I am developing a feature to score the test module based on the weightage given to test cases and applying mathematics to obtain the final score for the entire module. I need to print this score either in summary section or before the results  table.

j19sch:
resolves #109 by implementing pytest_html_results_summary hook

Hook allows adding data in a table in the Summary section. Styling is same as table in Environment section.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -424,8 +424,10 @@ class HTMLReport(object):
         body.extend(self._generate_environment(session.config))
 
         summary_prefix, summary_postfix = [], []
-        session.config.hook.pytest_html_results_summary(prefix=summary_prefix, summary=summary, postfix=summary_postfix)
-        body.extend([html.h2('Summary')] + summary_prefix + summary + summary_postfix)
+        session.config.hook.pytest_html_results_summary(
+            prefix=summary_prefix, summary=summary, postfix=summary_postfix)
+        body.extend([html.h2('Summary')] + summary_prefix
+                    + summary + summary_postfix)
 
         body.extend(results)
 
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -376,12 +376,12 @@ class HTMLReport(object):
         if self.rerun is not None:
             outcomes.append(Outcome('rerun', self.rerun))
 
-        summary = [html.h2('Summary'),
-                   self._generate_custom_summary(session.config),
-                   html.p('{0} tests ran in {1:.2f} seconds. '.format(
-                       numtests, suite_time_delta)),
-                   html.p('(Un)check the boxes to filter the results.',
-                          class_='filter', hidden='true')]
+        summary = [html.p(
+            '{0} tests ran in {1:.2f} seconds. '.format(
+                numtests, suite_time_delta)),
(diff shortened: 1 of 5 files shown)


pytest_html/plugin.py
```diff
@@ -376,12 +376,12 @@ def generate_summary_item(self):
...
-                numtests, suite_time_delta)),
-            html.p('(Un)check the boxes to filter the results.',
-                   class_='filter',
-                   hidden='true')]
+        summary = [html.h2('Summary'),
+                   self._generate_custom_summary(session.config),
+                   html.p('{0} tests ran in {1:.2f} seconds. '.format(
+                       numtests, suite_time_delta)),
+                   html.p('(Un)check the boxes to filter the results.',
+                          class_='filter', hidden='true')]
```
davehunt: It would be more powerful to call a `pytest_html_results_summary` with the complete summary list. The user can then decide to manipulate the list however they like. Alternatively, a `pytest_html_results_summary_prefix` or `pytest_html_results_summary_suffix` would preserve the default summary.

pytest_html/plugin.py
```diff
@@ -438,21 +438,37 @@ def generate_summary_item(self):
...
+    def _generate_custom_summary(self, config):
+        custom_summary = {}
+        config.hook.pytest_html_results_summary(custom_summary=custom_summary)
+
+        if not custom_summary:
+            return []
+        else:
+            return [html.table(
+                self._create_table_rows_from_dictionary(custom_summary),
+                id='custom-summary')]
```
davehunt: This limits any summary to being a dictionary that will be formatted as a table. I think this is unecessarily restrictive.

pytest_html/plugin.py
davehunt: Thinking about this further, we could have `pytest_html_results_summary` with three arguments: prefix, summary, and suffix. That would make it easy for the default summary to remain untouched, whilst allowing a prefix and suffix. For users that really want to, they can even manipulate the default summary.

j19sch: To make sure I understand you correctly:
- the `pytest_html_results_summary` hook will take three arguments: prefix, summary, suffix
- the argument type is a list of py.xml html objects
- the prefix by default contains the count/run time of the tests and the text about the filters (not sure about this one, alternative: default empty, always have count/run time and text)
- the summary by default contains the summary table
- the suffix by default is empty

pytest_html/plugin.py
j19sch: Good point, will change anyway when implementing the `pytest_html_results_summary` hook you suggested.

pytest_html/plugin.py
j19sch: @davehunt Please ignore previous comment. I implemented the hook with three arguments:
- `prefix`: empty by default
- `summary`: contains the count/run time of the tests and the text about the filters 
- `prefix`: empty by default

Type of all three arguments is a list of py.xml html objects.

PR #153, opened 2018-03-25
adds filename of report as title to html body

j19sch:
We have a CI dashboard with links to the reports generated for different tests. I'd like to see the filename of the report as title in the body, so it's clear which test results I'm looking at.

Note: I made the assumption that `os.path.basename` will work across platforms.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -414,6 +414,7 @@ class HTMLReport(object):
 
         body = html.body(
             html.script(raw(main_js)),
+            html.h1(os.path.basename(session.config.option.htmlpath)),
             html.p('Report generated on {0} at {1} by'.format(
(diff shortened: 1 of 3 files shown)


j19sch:
@davehunt You're welcome. :-)

About the `--title` command-line option: what about `--report-title` or `--html-title` to prevent collision with other plugins their command-line options? Not sure if there's some sort of convention for pytest plugins in this regard?

davehunt:
Of those options I'd probably go for `--html-title`, but thinking about it, perhaps we should hold back until someone asks for this feature.

Vic152:
I am asking for this feature :)

BeyondEvil:
Hi @Vic152 

Would you mind opening a separate ticket for this? Preferably with a clear use-case.

Thanks!

PR #162, opened 2018-05-03
Disable sort on environment table when metadata is ordered

### LINKED ISSUE #158: Custom environment table order
In Python 3.6+ or with `OrderedDict`, the config metadata can maintain its order to control the layout of the environment table. Would you consider an option to make `sorted()` optional in this line:

https://github.com/pytest-dev/pytest-html/blob/35a610065e50b94db30b068950e50d28f9c3545b/pytest_html/plugin.py#L468

jacebrowning:
Closes #158

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -96,6 +96,9 @@ via the :code:`pytest_configure` hook:
   def pytest_configure(config):
       config._metadata['foo'] = 'bar'
 
+The generated table will be sorted alphabetically unless the metadata is a
+:code:`collections.OrderedDict`.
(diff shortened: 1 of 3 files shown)


jacebrowning:
@davehunt Please take a look at this when you get a chance. 

I've already integrated this into one of much projects like so:

```
def pytest_configure(config):
    metadata = OrderedDict({
        'FOO': 1,
        'BAR': 2,
    })
    setattr(config, '_metadata', metadata)
``

PR #168, opened 2018-05-31
Customize collapsed rows with a query param

### LINKED ISSUE #157: feature request - open html log with closed details
Hi there,
We have adopted pytest-html to view our tests.
However, when a test fails it open the html file in a way we can't read it (it prints out the entire stacktrace + logs we are writing during the test).

It would be appropriate if we can pass down a value to close the html details of a failed/errored test and not opening it by default.

Thanks,
Mor

jacebrowning:
Closes #157 

The query param code is based on https://stackoverflow.com/a/5158301/429533.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,7 +12,7 @@ HTML report for the test results.
    :alt: PyPI
 .. image:: https://img.shields.io/conda/vn/conda-forge/pytest-html.svg
    :target: https://anaconda.org/conda-forge/pytest-html
-   :alt: Conda Forge   
+   :alt: Conda Forge
 .. image:: https://img.shields.io/travis/pytest-dev/pytest-html.svg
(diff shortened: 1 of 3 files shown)


jacebrowning:
> Expected: undefined, Actual: null

@davehunt Could you help me reason about these JavaScript failures?

davehunt:
Took me a while, but it appears that PhantomJS doesn't support `String.includes`... here's a solution from StackOverflow: https://stackoverflow.com/a/31361163/154975 I'd suggest adding this to `test.js`.

jacebrowning:
@davehunt Thanks! This should be ready for review.

davehunt:
Thanks @jacebrowning!

Formartha:
Great!!

PR #172, opened 2018-06-30
Fix invalid escape sequences

asottile:


asottile:
py36: silent warning
py37: warning
py38: syntax error

Simply preparing for the future (I fixed all of pytest-dev for this particular pattern)

davehunt:
I've merged this independently.

PR #178, opened 2018-09-17
Always define __version__ even if get_distribution() fails

nicoddemus:
In frozen executables created with tools such as pyinstaller, get_distribution() does not work because it looks for files in the environment which are not packed in the
executable.

This patch makes `__version__` available in those situations, otherwise the
plugin won't work at all because `plugin.py` tries to import `__version__`
from `pytest_html.__init__`.


nicoddemus:
Taking the liberty of merging Dave!

PR #181, opened 2018-10-02
Refactor css config code

crazymerlyn:
Change the default value for `--css` option to an empty array instead of None since the config reading code is doing that manually.


davehunt:
Thank you for your contribution @crazymerlyn

PR #189, opened 2019-01-01
adding English locale before datetime.now

avnaf:
previous pull request was setting locale too early and that change was overridden.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -38,6 +38,7 @@ if PY3:
 else:
     from codecs import open
     from cgi import escape
+    import locale
 
 
 def pytest_addhooks(pluginmanager):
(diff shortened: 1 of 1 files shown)


RibeiroAna:
Hi, just a small advice: You don't need to re-open your pull request, you just need to make a new commit to the same branch. I'll take a look on your pull request.

RibeiroAna:
Thank you for your PR. Could you add a test case on the file `/testing/test_pytest_html.py` showing us what kind of problem this PR solves? What I understood is that it allows unicode characters on test files, could you please create a test case with unicode characters?

BeyondEvil:
@avnaf Any progress on this? Please let us know if you need some assistance! :)

avnaf:
I have stopped working on it. It is a very small issue and was solved
locally

BeyondEvil:
@RibeiroAna Since you were the lead on this, I'll leave the decision to you whether to close it or not. :)

PR #190, opened 2019-01-04
Fixed tests that were failing due Pytest updates

RibeiroAna:
Comments explaining how I did it on the code.


.gitignore
```diff
@@ -12,6 +12,7 @@ dist
 # IDE Specific files
 ### Pycharm IDE - Jetbrains
 .idea/*
+.vscode/*
```
RibeiroAna: I'm using Visual Studio right now, so I added it.

pytest_html/plugin.py
```diff
@@ -113,8 +113,6 @@ def __init__(self, outcome, report, logfile, config):
             self.config = config
             self.row_table = self.row_extra = None
 
-            test_index = hasattr(report, 'rerun') and report.rerun + 1 or 0
```
RibeiroAna: I realised from [here](https://github.com/pytest-dev/pytest-rerunfailures/commit/4ec74311cd5b199a8ec6f198f17ea77ea16c56d6#diff-c3a581865b7c0be4e6b5d5fa60d6aed5) that report doesn't have the attribute `rerun` anymore, so that was causing some `rerun` tests to fail. Instead of getting the test_index from the attribute, I'm passing it as a parameter from the property `rerun` of the `TestResult` class.

testing/test_pytest_html.py
```diff
@@ -140,7 +140,9 @@ def test_xpass(): pass
 
     def test_setup_error(self, testdir):
         testdir.makepyfile("""
-            def pytest_funcarg__arg(request):
```
RibeiroAna: Funcarg is deprecated since [Pytest 4](https://github.com/pytest-dev/pytest/issues/1684).

testing/test_pytest_html.py
```diff
@@ -442,7 +444,7 @@ def pytest_runtest_makereport(item, call):
         """.format(extra_type, content))
         testdir.makepyfile("""
             import pytest
-            @pytest.mark.flaky(reruns=2)
+            @pytest.mark.flaky(reruns=4)
```
RibeiroAna: Just because the loop bellow goes from 1 to 4 (so it checks for extra content 4 times) and it was only 2 reruns.

pytest_html/plugin.py
```diff
@@ -263,7 +261,8 @@ def append_log_html(self, report, additional_html):
             additional_html.append(log)
 
     def _appendrow(self, outcome, report):
-        result = self.TestResult(outcome, report, self.logfile, self.config)
+        result = self.TestResult(outcome, report, self.logfile, self.config,
+                                 self.rerun)
```
davehunt: `self.rerun` will only be present if the pytest-rerunfailures plugin is installed, if it's not installed I would expect this to raise an exception due to the attribute not being available

pytest_html/plugin.py
davehunt: This could be a regression from pytest-rerunfailures. It would seem that the latest version doesn't report the number of reruns in the report? If so, please raise an issue against the plugin. In the meantime we could pin our tests to the version of pytest-rerunfailures before this change.

testing/test_pytest_html.py
davehunt: I see your logic, but I wonder why we had this mismatch with tests passing? This code hasn't changed in four years.

RibeiroAna: I saw the logic there before, actually the loop checks for content 3 times (`range(1,4) = [1, 2, 3]`, my fault) and the first test (the one which failed) count as the file1, the first rerun is actually  file2, and so on

tox.ini
```diff
@@ -10,7 +10,7 @@ envlist = py{27,36,37,py,py3}{,-ansi2html}, flake8
 commands = pytest -v -r a {posargs}
 deps =
     pytest-xdist
-    pytest-rerunfailures
+    pytest-rerunfailures <= 4.1
```
davehunt: Please add a comment referring to https://github.com/pytest-dev/pytest-rerunfailures/issues/77 so we can remove this in the future. Otherwise, this lgtm.

RibeiroAna:
Hi @davehunt Pytest-rerunfailures 6.0 has been released with my fix. Could you please approve this PR so we can have pytest-html with tests passing once again?

davehunt:
Sorry for the delay, and thanks @RibeiroAna!

PR #199, opened 2019-02-15
Change assets naming method

SunInJuly:
A little context:
I've been using pytest_html for layouts testing, sometimes reports contains about 100 mb of images, which is hard to navigate in browser. It would be nice if I could just scroll throw my assets and watch which test-case was failed in my file manager. 
I found out that assets naming contains exact info I need, but in some reason it hashed. 
So I added an option assets_name_hashing. When it's set False, assets names are not hashed and can be sorted nicely in folder. 

What do you think? 

Update: 
now all assets named like "test_name_[hash].jpg"

### DIFF
--- a/testing/test_pytest_html.py
+++ b/testing/test_pytest_html.py
@@ -383,12 +383,12 @@ class TestHTML:
         testdir.makepyfile('def test_pass(): pass')
         result, html = run(testdir)
         hash_key = ('test_extra_text_separated.py::'
-                    'test_pass00').encode('utf-8')
+                    'test_pass00')
         hash_generator = hashlib.md5()
-        hash_generator.update(hash_key)
+        hash_generator.update(hash_key.encode('utf-8'))
         assert result.ret == 0
-        src = '{0}/{1}'.format('assets', '{0}.txt'.
-                               format(hash_generator.hexdigest()))
+        src = '{0}/{1}'.format('assets', '{0}_{1}.txt'.
+                               format(hash_key, hash_generator.hexdigest()))
         link = ('<a class="text" href="{0}" target="_blank">'.format(src))
         assert link in html
         assert os.path.exists(src)
@@ -412,14 +412,12 @@ class TestHTML:
         """.format(extra_type, content))
         testdir.makepyfile('def test_pass(): pass')
         result, html = run(testdir)
-        hash_key = ('test_extra_image_separated.py::'
-                    'test_pass00').encode('utf-8')
+        hash_key = 'test_extra_image_separated.py::test_pass00'
         hash_generator = hashlib.md5()
-        hash_generator.update(hash_key)
+        hash_generator.update(hash_key.encode('utf-8'))
         assert result.ret == 0
-        src = '{0}/{1}'.format('assets', '{0}.{1}'.
-                               format(hash_generator.hexdigest(),
-                                      file_extension))
+        src = '{0}/{1}'.format('assets', '{0}_{1}.{2}'.
+                               format(hash_key, hash_generator.hexdigest(), file_extension))
         link = ('<a class="image" href="{0}" target="_blank">'.format(src))
         assert link in html
         assert os.path.exists(src)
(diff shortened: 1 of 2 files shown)


davehunt:
Without hashing you would expect the assets to be overwritten each time the tests are run. I'm fine with an option to prevent hashing, but I'm not a fan of lots of command line arguments. Perhaps this could be a config option or environment variable instead?

SunInJuly:
Ok, how about something like this?

RibeiroAna:
Hey @SunInJuly I think a good way of doing that is with Pytest-metadata, which is a plugin that you can set metadata info to your pytest (@davehunt tell me if I'm wrong).

But I think  I got a better ideia for this issue. What do you think of the name of the assets being the unhashed name plus the hash. Something like "test01_fail_testname_[hash].png", it would solve your issue plus not having the issue of overwriting the assets every time the tests are run.

SunInJuly:
Hey @RibeiroAna! 
Thanks for your idea, I will approach my issue this way! I just thought, there could be another reason for current naming that I don't understand. 
About Pytest-metadata - I read about it, but I don't undersatnd how is it different from what I am doing here? It would be super-helpful If you could give me links about it :)

RibeiroAna:
Pytest Metadata is kind of a new plugin (so no surprise that there isn't much about it on the web) that you can basically set some env variables to your test that can be read by pytest-html. So for example, you could set an option [like that](https://github.com/pytest-dev/pytest-metadata#accessing-metadata) and later access [like that](https://github.com/pytest-dev/pytest-html/blob/master/pytest_html/plugin.py#L473) in Pytest HTML.

BeyondEvil:
Hi @SunInJuly !

Instead of hard coding the default config value for hasing, I would add it, like @davehunt suggested, as an optional ini-value instead.

See [here](https://github.com/pytest-dev/pytest-selenium/blob/master/pytest_selenium/pytest_selenium.py#L374) and [here](https://github.com/pytest-dev/pytest-selenium/blob/master/pytest_selenium/pytest_selenium.py#L240)

More info [here](https://docs.pytest.org/en/latest/customize.html#command-line-options-and-configuration-file-settings)

RibeiroAna:
Thanks @BeyondEvil for the information! I was not so sure about my suggestion btw

davehunt:
> But I think I got a better ideia for this issue. What do you think of the name of the assets being the unhashed name plus the hash? Something like "test01_fail_testname_[hash].png", it would solve your issue plus not having the issue of overwriting the assets every time the tests are run.

I like this suggestion, it keeps the hashes but also makes the file names more descriptive, and the best thing is there's no new configuration to document! 😄

RibeiroAna:
I think there are enough tumbs up for my suggestion! So I think we won't need a configuration file, we can do just as I suggested. @SunInJuly could you update the PR? 😀

SunInJuly:
Yep! Here is what I came up with so far. 
I changed naming strategy and updated related tests. 
Naming works fine, BUT some tests seems to be broken :( 

Maybe you have tips for these tests? I'm working on it! 

FAILED testing/test_pytest_html.py::TestHTML::test_extra_image_separated_rerun[png-image]
FAILED testing/test_pytest_html.py::TestHTML::test_extra_image_separated_rerun[png-png]
FAILED testing/test_pytest_html.py::TestHTML::test_extra_image_separated_rerun[svg-svg]
FAILED testing/test_pytest_html.py::TestHTML::test_extra_image_separated_rerun[jpg-jpg]

SunInJuly:
Ok I got this! Tests are fixed!

testing/test_pytest_html.py
```diff
@@ -383,12 +383,12 @@ def pytest_runtest_makereport(item, call):
         testdir.makepyfile('def test_pass(): pass')
         result, html = run(testdir)
         hash_key = ('test_extra_text_separated.py::'
-                    'test_pass00').encode('utf-8')
+                    'test_pass00')
         hash_generator = hashlib.md5()
-        hash_generator.update(hash_key)
+        hash_generator.update(hash_key.encode('utf-8'))
```
RibeiroAna: I didn't get why you moved the econding from line 386 to here.

SunInJuly: I did the same thing in plugin.py. 
This way filename is just more clear. Without it filename would be something like: 
 b'test_name'_[hash].png

I wanted to remove useless "b" and single quotes

RibeiroAna: Gotcha! I think your PR is nice!

SunInJuly: wow, thank! ^_^

SunInJuly:
@RibeiroAna fixed it!

BeyondEvil:
Nice! Good job, @SunInJuly ! 👍

SunInJuly:
Thank you all, for being so helpful and supportative ^_^

PR #208, opened 2019-06-05
Allow opening generated html report in browser

ssbarnea:
By default terminals would open files in editors and URLs in browsers.

By printing the generated html report using an URL (file://) we enable
users to click the terminal to open it. Previously this opened the file
in the text editor which is very unlikely to be what user would want.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -521,5 +521,6 @@ class HTMLReport(object):
         self._save_report(report_content)
 
     def pytest_terminal_summary(self, terminalreporter):
-        terminalreporter.write_sep('-', 'generated html file: {0}'.format(
-            self.logfile))
+        terminalreporter.write_sep(
+            '-',
+            'generated html file: file://{0}'.format(self.logfile))


BeyondEvil:
> `file://foo` is not a correct file URL, see https://en.wikipedia.org/wiki/File_URI_scheme - also, you'd probably want to print the absolute filename there, I don't think relative file URLs are a thing.
> 
> At that point, I'd probably print both though, to make it easier to copy-paste the filename...

Actually, `file://` should work, since pytest-html already prints the absolute path: `/Users/jimbrannlund/dev/proxy/api/test/e2e/report.html`

ssbarnea:
Yep, the path is not relative by default which means that adding file:// prefix makes it valid. I tested and worked fine for me (final url has 3 slashes).

The-Compiler:
Ah, sorry, my bad! Didn't see that the path was absolute.

ssbarnea:
Any chance to see this merged and ideally included in a release? I kinda miss it.

ssbarnea:
@RibeiroAna Thanks!

PR #217, opened 2019-07-09
Append a line break after captured log sections

borntyping:
The content in log sections doesn't always include a trailing newline, so this includes one afterwards ensuring that the next section header is placed on a new line.

This is only really useful if there are multiple log sections, like in the example below. If there's already a trailing newline, this has no effect as it adds a `<br>` tag which doesn't change the spacing at the end of the content.

<details>
<summary>Example</summary>

```python
import pytest


@pytest.fixture()
def logger():
    import logging

    log = logging.getLogger(__name__)
    log.info("before the test")
    return log


def test_log(logger):
    logger.info("during the test")
    raise Exception
```

</details>

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -263,6 +263,7 @@ class HTMLReport(object):
                     converter = Ansi2HTMLConverter(inline=False, escaped=False)
                     content = converter.convert(content, full=False)
                 log.append(raw(content))
+                log.append(html.br())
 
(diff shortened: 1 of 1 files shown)


BeyondEvil:
Thank you for this @borntyping ! 

This has annoyed me for ever, but I've never gotten around to actually fixing it.

PR #218, opened 2019-07-10
Migrate to 'black' for consistent styling and formatting

BeyondEvil:
[black](https://github.com/python/black)


BeyondEvil:
Sorry to bother you @nicoddemus, but I think I need your help.

When running tests using xdist, some tests [fail](https://travis-ci.org/pytest-dev/pytest-html/builds/556974278) when using 2.7 and pypy.

But running without xdist they [pass](https://travis-ci.org/pytest-dev/pytest-html/builds/556980302).

Any idea why that is, and if it's fixable?

Thanks! 🙏

nicoddemus:
> Sorry to bother you @nicoddemus, but I think I need your help.

No problem at all!

I did some investigation and I suspect that it is related to a long-standing issue on how workers are initialized in xdist, see https://github.com/pytest-dev/pytest-xdist/pull/448. I'm afraid there's no simple solution for now, because https://github.com/pytest-dev/pytest-xdist/pull/448 will require pytest>=5.1, which no longer supports Python 2.7 anyway. 😕 

My suggestion is to just execute `pytest-html` without xdist for now on Python 2.

BeyondEvil:
Thank you @nicoddemus ! I appreciate it!

Glad I didn't go down the rabbit hole of trying to troubleshoot this then. I'll do what you advise. 😊 

Thanks again!

Issue #219, opened 2019-07-11
pytest html does not keep sort preference for previously sorted columns


wanam:
Let's take this example to show this columns sort issue on the output html report:
**Column1 Column2**
C1_T2    C2_T1
C1_T2    C2_T2
C1_T4    C2_T4
C1_T1    C2_T5

Here we have the column2 that is sorted asc, if we click on the header Column1 to sort asc here is what we will get:
**Column1 Column2**
C1_T1    C2_T5
C1_T2    C2_T2
C1_T2    C2_T1
C1_T4    C2_T4

As you see it sorted the first column "asc" as requested, BUT it reversed the Column2 sort to "desc":
C1_T2    **C2_T2**
C1_T2    **C2_T1**

Here is the expected behavior:
**Column1 Column2**
C1_T1    C2_T5
C1_T2    C2_T1
C1_T2    C2_T2
C1_T4    C2_T4

PR #220, opened 2019-07-11
Keep sort preference for previously sorted columns

wanam:
This fixes https://github.com/pytest-dev/pytest-html/issues/219


BeyondEvil:
Thanks @wanam 

Could you add a test for this as well?

wanam:
I added 2 assertions to check this issue on `sort_column` test.

BeyondEvil:
Could you update with master and fix failing tests @wanam ?

wanam:
@BeyondEvil I'm getting some troubles trying to install your CI environment because of a grunt-contrib-qunit's dependency 'puppeteer' that fails to download Chromium module, maybe because I'm behind a proxy...

Anyway executing the tests on Chrome browser gave me the same result with the 3 failed tests, this cannot reproduced on my default browser: Firefox, do you have any idea what could be the reason for this?

BeyondEvil:
js tests and related issues are not my fortay. Maybe @RibeiroAna has some input? 

If not, I'll see if I can illicit some help from a colleague. 🤷‍♂

wanam:
I still don't understand why we have different behavior on Firefox vs Chrome, but I just pushed a fix for the sorting issue, the root cause was in case of equal keys, for this case the sort function should return 0.

@BeyondEvil If it's ok to merge, could you squash my commits into one commit?

PR #224, opened 2019-08-05
Remove unused and undocumented markers

BeyondEvil:
Seems like ´optionalhook´ was never used.

´hookwrapper´ generated `PytestUnknownMarkWarning`

and I couldn't find anything regarding it explicitly in the changelog of `pytest`.

Replaced it with `@pytest.hookimpl(hookwrapper=True)` which did not generate any warnings.

Care to shed some light on the above @nicoddemus @RonnyPfannschmidt ?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,7 +112,6 @@ You can edit the *Summary* section by using the :code:`pytest_html_results_summa
    import pytest
    from py.xml import html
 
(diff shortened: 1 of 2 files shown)


nicoddemus:
@BeyondEvil not sure, doesn't ring a bell at the moment, but the changes LGTM.

PR #235, opened 2019-09-12
Close opened resource.

krzysztof-pawlik-gat:
When using custom CSS file I'm getting following warning:

```
/home/nelchael/.local/share/virtualenvs/functional-tests-IjeTcJ7i/lib/python3.7/site-packages/pytest_html/plugin.py:71: ResourceWarning: unclosed file <_io.TextIOWrapper name='resources/pytest-html-custom.css' mode='r' encoding='UTF-8'>
  open(csspath)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

This fixes it by closing the opened file. If I'm understanding it correctly, the file is opened just to make sure it exists here, so it could be potentially replaced with `os.access` call, but it's entirely optional.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -69,7 +69,7 @@ def pytest_configure(config):
     if htmlpath:
         for csspath in config.getoption("css"):
             if not os.path.exists(csspath):
-                raise IOError("No such file or directory: '{}'".format(csspath))
+                raise IOError(f"No such file or directory: '{csspath}'")
         if not hasattr(config, "slaveinput"):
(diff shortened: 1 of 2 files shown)


krzysztof-pawlik-gat:
> It is possible to add a test for this? Essentially checking that we get the warning prior to the fix, and don't get it after the fix? @krzysztof-pawlik-gat

Sorry for the late response, but yes and it's actually quite easy - added modifications to tests in 05ac9a8 (checked with the fix reverted and tests fail due to warnings, with the fix tests pass as expected).

PR #237, opened 2019-09-12
Fix min-width which produced horizontal scrolls

ssbarnea:
Previous min-width caused undesired horizontal scrolls even on a 5K
display with 1/2 split screen. Reduced value allows users to better
use their screens.

### DIFF
--- a/pytest_html/resources/style.css
+++ b/pytest_html/resources/style.css
@@ -1,7 +1,8 @@
 body {
 	font-family: Helvetica, Arial, sans-serif;
 	font-size: 12px;
-	min-width: 1200px;
+	/* do not increase min-width as some may use split screens */
+	min-width: 800px;
 	color: #999;
 }
 


BeyondEvil:
Would you mind checking this out @rochacbruno please?

I'm not particularly well versed in CSS, but after having had a quick look - this feels like we're solving a symptom instead of a problem.

Thanks! 🙏

ssbarnea:
When it comes to CSS there is no perfect solution. In the end there must be a minimal width. I doubt people would be able to read anyoning with a 1px min width. Also having a too big value causes the problem I described. Clearly 1200 is too much, I am not sure if 800 perfect but clearly is better than 1200. ;)

BeyondEvil:
Yeah, you're probably right @ssbarnea !

This plugin has been around a while now, almost 5 years, and haven't seen much love when it comes to the esthetics.

I wish I had more time, and more knowledge specific to HTML and CSS. I'd like to do a bigger refactor.

PR #250, opened 2019-11-05
Release v2.0.1

BeyondEvil:


.travis.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         directories:
           - $HOME/.cache/pre-commit
 
+    -
```
BeyondEvil: This caused py36 to be run instead of linting...

testing/test_pytest_html.py
```diff
@@ -786,16 +786,16 @@ def test_css_invalid_no_html(self, testdir):
...
-                assert eval(input) == expected
+            @pytest.mark.parametrize("utf8", [("测试用例名称")])
+            def test_pass(utf8):
+                assert True
         """
         )
         result, html = run(testdir)
         assert result.ret == 0
-        assert r'\u6d4b\u8bd5\u7528\u4f8b\u540d\u79f0' not in html
+        assert r"\u6d4b\u8bd5\u7528\u4f8b\u540d\u79f0" not in html
```
BeyondEvil: ...which is why most of these changes came to be.

Issue #245, opened 2019-10-16
Dictionaries in environment table a not sorted


ssbarnea:
As in Python dictionaries are not (usually) sorted printing the environment gives random results, making much harder to read it and to compare it with previous runs.

Example: http://logs.rdoproject.org/52/23152/1/check/rdo-tox-molecule/6eb5f91/tox/reports.html

BeyondEvil:
Fair point!

Feel free to contribute with a fix! 😄 

There's a couple of other issues with your name on them 😉

ssbarnea:
@BeyondEvil Sad to dissapoint you but since you dropped support for py27 I lost my interest of contributing as most of our jobs are still on python27. Hopefully this may change next year, once I manage to migrate more jobs.

While I support dropping support for py27, pytest stuff is not one of the places to start because is used for testing, lots of people will continue to use it next year because they have LTS maintenance branches with CI/CD on them. When you cannot fix the versions you are using the incentive to contribute goes down rapidly.

BeyondEvil:
@ssbarnea I see, that is disappointing indeed.

And I guess, it's our fault for not communicating properly, but we haven't dropped support for python 2.7 (at least not the way you think we have).

Read more here: https://docs.pytest.org/en/latest/py27-py34-deprecation.html

PR #260, opened 2020-02-02
Video support for mp4 format

ExaltedBagel:
# Summary
This feature enables the use of extra videos in the reports in the same way as images are used.
```
extra.video(content, name="Video", mime_type="video/mp4", extension="mp4")
extra.mp4(content, name="Video")
```
## Details
- Videos may be included as reference, or as standalone
- Both of these cases are managed as the images are. Base64 encoding is used to pack the video as standalone data if needed.
- Extra video and windows path for videos are both added to the unit tests.
- Video and images were refactored to use a common media function internally to eliminate code duplication.

### DIFF
--- a/pytest_html/extras.py
+++ b/pytest_html/extras.py
@@ -7,6 +7,7 @@ FORMAT_IMAGE = "image"
 FORMAT_JSON = "json"
 FORMAT_TEXT = "text"
 FORMAT_URL = "url"
+FORMAT_VIDEO = "video"
 
 
 def extra(content, format, name=None, mime_type=None, extension=None):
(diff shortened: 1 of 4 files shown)


ExaltedBagel:
I think I would be done for this feature. Feel free to merge whenever you are ready! I might grab another issue in the meantime such as #245

robbiewareham:
This is awesome! Any date on when the new release will be built?

BeyondEvil:
I'm aiming to do a release during the weekend @robbiewareham

BeyondEvil:
**Released in v2.1.0**

BeyondEvil:
Hey @ExaltedBagel !

I discovered a bug with your PR.

On this [line](https://github.com/pytest-dev/pytest-html/blob/v2.0.1/pytest_html/plugin.py#L238), `href` will always be `none` in the case of `FORMAT_IMAGE`.

Just wanted you to know, and I might ask you to review the fix. 👍

PR #262, opened 2020-02-05
Added support for ordering metadata dictionaries by key.

### LINKED ISSUE #245: Dictionaries in environment table a not sorted

ExaltedBagel:
# Summary
This pull request fixes #245 by sorting environment metadata dictionaries by keys.

## Limitations

- This fix does not implement any form of recursion. This means that if you have a metadata value which has a dictionnary value, this inner dictionary will not be sorted.
- In order to be able to compare the dictionary key values, they are cast as string. This allows a lexical sorting order, which is a natural way of sorting.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -510,7 +510,8 @@ class HTMLReport:
                 key_value_list = [f"'{k}': {value[k]}" for k in sorted(value)]
                 value = ", ".join(key_value_list)
                 value = "{" + value + "}"
-            rows.append(html.tr(html.td(key), html.td(raw(value))))
+            raw_value_string = raw(str(value))
+            rows.append(html.tr(html.td(key), html.td(raw_value_string)))
 
         environment.append(html.table(rows, id="environment"))
         return environment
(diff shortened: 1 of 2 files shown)


ExaltedBagel:
Seems ready to go. Tell me if you would like a greater test coverage.

BeyondEvil:
Nice! Looks really good, thanks for this!

Greater test coverage would be greatly appreciated. For instance, will it handle metadata that is script/html?

As for the limitations, I think those are totally fine. Easy enough to extend your work if the need should arise.

Please rebase with master (I merged your other PR 🙏 ).

ExaltedBagel:
The test for the dictionary containing some html has been done. It seems to be behaving as expected. I cleaned up the code a bit too. If your review is positive, I would be ready to merge!

BeyondEvil:
**Released in v2.1.0**

Issue #239, opened 2019-09-17
Opening report is slow when it contains long logs


Wramberg:
I have some test runs that generate a lot of log entries. When opening the report is seems my browser is spending a lot of time presenting all these entries which are immediately collapsed right after.

If I change the behavior to have everything collapsed by default I can open/render the html report much faster.

What I have tested is essentially the same as adding a "collapsed" class here https://github.com/pytest-dev/pytest-html/blob/master/pytest_html/plugin.py#L140

I have no idea if this breaks other use-cases or if other changes are necessary. I also realize that everything related to the collapse functionality is currently in js. Regardless, if something could be done to generate everything collapsed and open the necessary stuff afterwards I believe performance would be a lot better.

Edit: I should mention I use self-contained reports.

fpiccione:
@Wramberg maybe [this](https://stackoverflow.com/questions/57282301/send-html-report-with-row-collapsed) will work ?

Wramberg:
@fpiccione Yes it will. But my issue is about having to make a manual hack.

BeyondEvil:
Would this help: https://github.com/pytest-dev/pytest-html/pull/168 @Wramberg

Wramberg:
@BeyondEvil I dont think so. It seems to change which rows are collapsed but not the fact that ALL rows as rendered as non-collapsed just to be collapsed later by javascript. If we instead rendered the passed rows collapsed the rendering would be a lot faster.

BeyondEvil:
@Wramberg Ah, I see.

How would you like to see this accomplished? INI-file (pytest.ini etc) setting?

Wramberg:
I imagine it can simply be changed permanently. I do not see this breaking anything since it would not change the final rendered page, just the way it is rendered - but perhaps I am missing something ?

Regardless a command line option or INI setting would be fine for me.

BeyondEvil:
Even tho it wouldn't technically break something, I'm reluctant to change a behaviour going back multiple years.

Happy to add an INI setting however.

Wramberg:
I understand. Whichever approach you prefer :+1:

BeyondEvil:
Hey @Wramberg 

Would you mind perf-testing my PR?

Wramberg:
@BeyondEvil Sure. I will check it out tomorrow.

BeyondEvil:
I merged it. But please feel free to use master branch. Good to have a test before I do a release. @Wramberg

[...]

PR #267, opened 2020-02-15
Control render using configuration

BeyondEvil:
Fixes: #239

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -240,9 +240,16 @@ Display options
 
 By default, all rows in the **Results** table will be expanded except those that have :code:`Passed`.
 
-This behavior can be customized with a query parameter: :code:`?collapsed=Passed,XFailed,Skipped`.
+This behavior can be customized either with a query parameter: :code:`?collapsed=Passed,XFailed,Skipped`
+or by setting the :code:`render_collapsed` in a configuration file (pytest.ini, setup.cfg, etc).
 
+.. code-block:: ini
 
+  [pytest]
+  render_collapsed = True
+
+**NOTE:** Setting :code:`render_collapsed` will, unlike the query parameter, affect all statuses.
+
 Screenshots
 -----------
 
--- a/README.rst
+++ b/README.rst
@@ -246,10 +246,10 @@ or by setting the :code:`render_collapsed` in a configuration file (pytest.ini, 
 .. code-block:: ini
 
   [pytest]
-  render_collapsed = True
+  render_collapsed = True
 
 **NOTE:** Setting :code:`render_collapsed` will, unlike the query parameter, affect all statuses.
-
+
 Screenshots
 -----------
 
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -62,6 +62,12 @@ def pytest_addoption(parser):
         default=[],
         help="append given css file content to report style file.",
     )
+    parser.addini(
+        "render_collapsed",
+        type="bool",
+        default=False,
+        help="Open the report with all rows collapsed. Useful for very large reports",
+    )
 
 
 def pytest_configure(config):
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -240,9 +240,16 @@ Display options
...
 By default, all rows in the **Results** table will be expanded except those that have :code:`Passed`.
 
-This behavior can be customized with a query parameter: :code:`?collapsed=Passed,XFailed,Skipped`.
+This behavior can be customized either with a query parameter: :code:`?collapsed=Passed,XFailed,Skipped`
+or by setting the :code:`render_collapsed` in a configuration file (pytest.ini, setup.cfg, etc).
 
+.. code-block:: ini
 
+  [pytest]
+  render_collapsed = True
```
nicoddemus: Is there an extra character after `True` here?

README.rst
```diff
@@ -240,9 +240,16 @@ Display options
...

+
+**NOTE:** Setting :code:`render_collapsed` will, unlike the query parameter, affect all statuses.
+
```
nicoddemus: Here too

README.rst
BeyondEvil: Damn, you're right. 🤔 

1. How did you see them?
2. Any idea where they come from? They weren't visible characters, just some weird ASCII(maybe?).

nicoddemus: No idea :(

BeyondEvil: How did you notice/see them? I can't tell in the Github diff.

nicoddemus: Weird they don't show up anymore... 🤔

BeyondEvil: I fixed them @nicoddemus :) (the second commit)

nicoddemus: Great, that's fixed then

Wramberg:
@BeyondEvil As agreed in #239 I gave this a shot. The render is must faster now but unfortunately the collapsed class is misplaced. This means it is no longer possible to unfold the details/extras. With this PR, setting render_collapsed to true yields
```html
<tr>
  <td class="extra collapsed" colspan="4">
    <div>
```
but it should be
```html
<tr class="collapsed">
  <td class="extra" colspan="4">
    <div>
```
By changing the implementation to the following it works for me
```python
            if len(cells) > 0:
                tr_class = None
                if self.config.getini("render_collapsed"):
                    tr_class = "collapsed"

                self.row_table = html.tr(cells)
                self.row_extra = html.tr(
                    html.td(self.additional_html, class_="extra", colspan=len(cells)), class_=tr_class
                )
```
Essentially just moving the collapsed class to the parent `html.tr()`.

BeyondEvil:
> @BeyondEvil As agreed in #239 I gave this a shot. The render is must faster now but unfortunately the collapsed class is misplaced. This means it is no longer possible to unfold the details/extras. With this PR, setting `render_collapsed=true` yields
> 
> ```
> <tr>
>   <td class="extra collapsed" colspan="4">
>     <div>
> ```
> 
> but it should be
> 
> ```
> <tr class="collapsed">
>   <td class="extra" colspan="4">
>     <div>
> ```
> 
> By changing the implementation to the following it works for me
> 
> ```python
>             if len(cells) > 0:
>                 tr_class = None
>                 if self.config.getini("render_collapsed"):
>                     tr_class = "collapsed"
> 
>                 self.row_table = html.tr(cells)
>                 self.row_extra = html.tr(
>                     html.td(self.additional_html, class_="extra", colspan=len(cells)), class_=tr_class
>                 )
> ```
> 
> Essentially just moving the collapsed class to the parent `html.tr()`.

lol, awkward.

Good catch! In the end it was good, because I noticed JS needed an update as well to display the correct CSS.

BeyondEvil:
**Released in v2.1.0**

PR #268, opened 2020-02-20
Fix collapsable bug

### LINKED ISSUE #239: Opening report is slow when it contains long logs

BeyondEvil:
Fixes #239 (again)

Had placed the `collapsed` class in the wrong tag.

Which in the end was a good thing since I discovered that JS needed an update as well to display the correct CSS (show/hide text).

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -143,12 +143,13 @@ class HTMLReport:
             )
 
             if len(cells) > 0:
-                td_class = "extra"
+                tr_class = None
                 if self.config.getini("render_collapsed"):
-                    td_class += " collapsed"
+                    tr_class = "collapsed"
                 self.row_table = html.tr(cells)
                 self.row_extra = html.tr(
-                    html.td(self.additional_html, class_=td_class, colspan=len(cells))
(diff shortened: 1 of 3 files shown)


BeyondEvil:
ping reviewers @nicoddemus @rochacbruno @davehunt

nicoddemus:
Thanks for the ping @BeyondEvil, but I'm not sure I can review this PR right now. Others are welmcome to!

BeyondEvil:
**Released in v2.1.0**

PR #270, opened 2020-02-26
Add hook to change report main title

werdeil:
I think it is a good thing to be able to have a customizable report title that is not always the same as the file name.

As it is my first merge request on this project I'm open to any comment.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -99,6 +99,7 @@ class HTMLReport:
         logfile = os.path.expanduser(os.path.expandvars(logfile))
         self.logfile = os.path.abspath(logfile)
         self.test_logs = []
+        self.title = os.path.basename(self.logfile)
         self.results = []
         self.errors = self.failed = 0
         self.passed = self.skipped = 0
@@ -490,9 +491,11 @@ class HTMLReport:
             __name__, os.path.join("resources", "main.js")
         ).decode("utf-8")
 
+        session.config.hook.pytest_html_report_title(report=self)
+
         body = html.body(
             html.script(raw(main_js)),
-            html.h1(os.path.basename(self.logfile)),
+            html.h1(self.title),
             html.p(
                 "Report generated on {} at {} by ".format(
                     generated.strftime("%d-%b-%Y"), generated.strftime("%H:%M:%S")
--- a/README.rst
+++ b/README.rst
@@ -87,6 +87,19 @@ be used to change the appearance of the report.
 
   $ pytest --html=report.html --css=highcontrast.css --css=accessible.css
 
+Report Title
+~~~~~~~~~~~~
+
+By default report title will be the filename of the report, you can edit it by using the :code: `pytest_html_report_title` hook:
+
+.. code-block:: python
+
+   import pytest
+   from py.xml import html
+
+   def pytest_html_report_title(report)
+      report.title = "My very own title!"
+
 Environment
 ~~~~~~~~~~~
 
--- a/pytest_html/hooks.py
+++ b/pytest_html/hooks.py
@@ -3,6 +3,10 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
+def pytest_html_report_title(report):
+    """ Called before adding the title to the report """
+
+
 def pytest_html_results_summary(prefix, summary, postfix):
     """ Called before adding the summary section to the report """
 
--- a/testing/test_pytest_html.py
+++ b/testing/test_pytest_html.py
@@ -904,3 +904,19 @@ class TestHTML:
         assert result.ret == 1
         assert len(re.findall(collapsed_html, html)) == expected_count
         assert_results(html, tests=2, passed=1, failed=1)
+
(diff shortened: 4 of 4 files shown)


BeyondEvil:
Thanks for this PR @werdeil !

Couple of comments:

- Tests on Travis are failing. Specifically linting. Please see [this](https://github.com/pytest-dev/pytest-html/blob/master/development.rst) on how to fix.
- Please do add some documentation regarding the hook [here](https://github.com/pytest-dev/pytest-html/blob/master/README.rst), bonus points if you add an example.
- Please also add some tests to cover the new functionality.

Feel free to reach out if you need any guidance.

Thanks again!

pytest_html/plugin.py
```diff
@@ -490,9 +490,12 @@ def generate_summary_item(self):
             __name__, os.path.join("resources", "main.js")
         ).decode("utf-8")
 
+        report_title = os.path.basename(self.logfile)
+        session.config.hook.pytest_html_results_report_title(report_title=report_title)
```
csm10495: Isn't the behavior for this a bit confusing? I thought strings were immutable.

```
>>> def swap(rt):
...     rt = 'hi'
...
>>> rt = 'report title'
>>> swap(rt)
>>> rt
'report title'
>>>            
```

How can the hook modify the report_title in place? Shouldn't the hook return a value instead?

werdeil: Indeed, did think it about enough, sorry.

I see two solutions:
- the hook return a value but it doesn't work as the other hooks.
- the hook takes a mutable as input and we keep this behavior

I think the second option is better, and I'll push a commit to give the whole html title to the hook. It will allow more title customization.
Just need to document it, add test and examples before submitting again.

werdeil:
I think I've covered all review comments, Travis tests are failing but I think this is not related to my changes, I would like to rerun it but I don't have the right to do it.

@BeyondEvil @csm10495 Tell me if I need to do anything more.

BeyondEvil:
Is there a specific use-case for the hook @werdeil ?

With the way it is implemented, it's a lot of code (perhaps even the introduction of a conftest file) just to set the report title.

If there's not a specific use-case that warrants a hook, maybe a command line argument and/or an ini-file key will suffice? See this [comment](https://github.com/pytest-dev/pytest-html/pull/153#pullrequestreview-106951069)

werdeil:
My use case is to have a CI pipeline that generates a customized report.html. The CI generates automatically the title, and may add dynamically version or number of pass/fail e.g. "TEST REPORT v1.2.3 [2/3 Passed]"

In my use case I already have a conftest, and use other hooks so having another hook seems logic. 

Having a command-line and/or an ini file key will indeed be sufficient except in cases where we want to add some test results in it.
Personally I think it is better to have it as a hook (even if means having a conftest) as I see it similar as other report.html customizations (all customizations in the same file, but tell me if you want me to explore a solution with the `--html-title` option.

BeyondEvil:
Ok, fair enough! Thanks for the detailed explanation.

So, I guess with that, what I would like to see is to make it a little more straight forward to set the title than to have to mess around with html etc.

Here's my suggestion:

Add a `title` property to the [HTMLReport](https://github.com/pytest-dev/pytest-html/blob/master/pytest_html/plugin.py#L97) class.

Then you can make the hook like so:

```python
def pytest_html_report_title(report):
    """ Called before adding the title to the report """
```

And the plugin can call the hook:

```python
session.config.hook.pytest_html_report_title(report=self)
```

And an implementation can then be made like:

```python
def pytest_html_report_title(report):
    report.title = "My very own title!"
```

Also, _note_ the name change of the hook. The title is for the entire report, not only the results part. 😊 

Hope that makes sense. Feel free to question it. :)

werdeil:
Ok I'll follow your suggestion but I have one remark though: as we give the report to the hook, it allows to change all parts of the report, not only the title.

I see that we have already hooks that take the report as input so I guess it is acceptable. My concern is that someone willing to change anything in the report will use the "pytest_html_report_title" hook instead of finding a "better" way to do it, but as I said, it is already possible though other hooks.

pytest_html/plugin.py
```diff
@@ -99,6 +99,7 @@ def __init__(self, logfile, config):
         logfile = os.path.expanduser(os.path.expandvars(logfile))
         self.logfile = os.path.abspath(logfile)
         self.test_logs = []
+        self.title = ""
```
BeyondEvil:
```suggestion
        self.title = os.path.basename(self.logfile)
```

pytest_html/plugin.py
```diff
@@ -490,9 +491,12 @@ def generate_summary_item(self):
             __name__, os.path.join("resources", "main.js")
         ).decode("utf-8")
 
+        self.title = os.path.basename(self.logfile)
```
BeyondEvil: With the default above (the suggestion), you can remove this line.

werdeil: Done

BeyondEvil:
**Released in v2.1.0**

PR #271, opened 2020-02-28
Post process html to include teardown in log

### LINKED ISSUE #131: Report fixture setup and teardown
Hi,
I would like to vizualize when fixtures are set up and teared down, the time they take, their output and if they failed.
I think they could be added in the Results section, filtered by default, around the tests.
It might add a lot of information, so it would look better along with #33 

Would you be open to such an enchancement? (It could be optional if needed.)
Do you see issues it might create? Would you do it in a different way?

csm10495:
I believe this would resolve #131 ... however it isn't really optimal, I didn't want to dig into the html logic, so instead just moved it to be post processed and tried to coerce the test result into what the existing code seemed to want.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -11,6 +11,7 @@ import time
 import warnings
 from base64 import b64decode
 from base64 import b64encode
+from collections import defaultdict
 from collections import OrderedDict
 from functools import lru_cache
 from html import escape
@@ -148,6 +149,7 @@ class HTMLReport:
         self.rerun = 0 if has_rerun else None
         self.self_contained = config.getoption("self_contained_html")
         self.config = config
+        self.reports = defaultdict(list)
 
     class TestResult:
         def __init__(self, outcome, report, logfile, config):
@@ -279,7 +281,12 @@ class HTMLReport:
         def append_log_html(self, report, additional_html):
             log = html.div(class_="log")
             if report.longrepr:
-                for line in report.longreprtext.splitlines():
+                # longreprtext is only filled out on failure by pytest
+                #    otherwise will be None.
+                #  Use full_text if longreprtext is None-ish
+                #   we added full_text elsewhere in this file.
+                text = report.longreprtext or report.full_text
+                for line in text.splitlines():
                     separator = line.startswith("_ " * 10)
                     if separator:
                         log.append(line[:80])
@@ -620,15 +627,66 @@ class HTMLReport:
             with open(style_path, "w", encoding="utf-8") as f:
                 f.write(self.style_css)
 
+    def _post_process_reports(self):
+        for test_name, test_reports in self.reports.items():
+            outcome = "passed"
+            wasxfail = False
+            failure_when = None
+            full_text = ""
+            extras = []
+            duration = 0.0
+
+            # in theory the last one should have all logs so we just go
+            #  through them all to figure out the outcome, xfail, duration,
+            #    extras, and when it swapped from pass
+            for test_report in test_reports:
+                full_text += test_report.longreprtext
+                extras.extend(getattr(test_report, "extra", []))
+                duration += getattr(test_report, "duration", 0.0)
+
+                if (
+                    test_report.outcome not in ("passed", "rerun")
+                    and outcome == "passed"
+                ):
+                    outcome = test_report.outcome
+                    failure_when = test_report.when
+
+                if hasattr(test_report, "wasxfail"):
+                    wasxfail = True
+
+                if test_report.outcome == "rerun":
+                    self.append_other(test_report)
+
+            # the following test_report.<X> = settings come at the end of us
+            #  looping through all test_reports that make up a single
+            #    case.
+
+            # outcome on the right comes from the outcome of the various
+            #  test_reports that make up this test case
+            #    we are just carrying it over to the final report.
+            test_report.outcome = outcome
+            test_report.when = "call"
+            test_report.nodeid = test_name
+            test_report.longrepr = full_text
+            test_report.extra = extras
+            test_report.duration = duration
+
+            if wasxfail:
+                test_report.wasxfail = True
+
+            if test_report.outcome == "passed":
+                self.append_passed(test_report)
+            elif test_report.outcome == "skipped":
+                self.append_skipped(test_report)
+            elif test_report.outcome == "failed":
+                test_report.when = failure_when
+                self.append_failed(test_report)
+
+            # we don't append other here since the only case supported
+            #  for append_other is rerun, which is handled in the loop above
+
     def pytest_runtest_logreport(self, report):
-        if report.passed:
-            self.append_passed(report)
-        elif report.failed:
-            self.append_failed(report)
-        elif report.skipped:
-            self.append_skipped(report)
-        else:
-            self.append_other(report)
+        self.reports[report.nodeid].append(report)
 
     def pytest_collectreport(self, report):
         if report.failed:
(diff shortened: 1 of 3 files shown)


csm10495:
Wow. Those unit tests really helped me fix a lot of things i had broken :P. 

Honestly don't know if this is good enough to merge... it seems kind of sketchy... but per the tests seems to work. :)

BeyondEvil:
Thanks for this! @csm10495 

I need some time to fully understand the problem _and_ the fix before I can review this with confidence.

Feel free to ping me if you haven't heard anything from me in a week or so.

csm10495:
Sure thing @BeyondEvil . The big thing that I want to fix is to get the teardown event in the html log:

for example:


```
import pytest
pytest.fixture(scope='function')
def setupAndTeardown():
    print ("setup: i am already in the html")
    yield
    print ("teardown: i am missing from the html... but added by this change!")

def test_nothing():
    print ("test_case: of course i am already in the html")
```

In doing that, i try to merge hold back from processing the reports, until we get to the ```pytest_sessionfinish()``` then basically combine the setup/testcase/teardown reports to figure out the status, etc.

csm10495:
@BeyondEvil reminder to check this out. Thanks.

csm10495:
Reminder to check this out @BeyondEvil

dasm:
@csm10495 @BeyondEvil I'd like to help with that functionality. Do you need anything extra?

Interesting fact. I'm able to see `captured stdout teardown` in HTML if teardown failed.

Formartha:
Hi folks, any update about this one? we're waiting for it as well :).
a W/A can be provided in case you have one till it will be merged..

dina809:
Hi @csm10495 - i have tried your code and it works good while collecting logs from teardown of a fixture and showing them in the html report,
but it is causing an issue if the fixture setup is failing (before the yield), have you checked that?

csm10495:
Hey @dina809 , it's been a while now since I've looked deep at this. What error are you seeing? It works if I do this:

```
import pytest
@pytest.fixture(scope="function")
def setupAndTeardown():
    print ("this is setup")
    assert 0
    yield
    print ("this is teardown")
def test_setup_and_teardown(setupAndTeardown):
    print ("this is the test case")
```

HTML:
image

avzero07:
Bump.

1Mark:
Any ETA for release date for this?

BeyondEvil:
> Any release date for this?

I'm hoping @ssbarnea  or @gnikonorov can find some time to review this?

gnikonorov:
I can take a look at this tonight, @BeyondEvil

pytest_html/plugin.py
```diff
@@ -273,7 +275,8 @@ def append_extra_html(self, extra, extra_index, test_index):
         def append_log_html(self, report, additional_html):
             log = html.div(class_="log")
             if report.longrepr:
-                for line in report.longreprtext.splitlines():
+                text = report.longreprtext or report.full_text
```
gnikonorov, line 278: Can you add a comment explaining why `longreprtext` is Falsey at times and why we default to `full_text` in those instances?

pytest_html/plugin.py
```diff
@@ -612,15 +615,57 @@ def _save_report(self, report_content):
...
+            failure_when = None
+            full_text = ""
+            extras = []
+            duration = 0.0
+
+            # in theory the last one should have all logs so we just go
+            #  through them all to figure out the outcome, xfail, duration,
+            #    extras, and when it swapped from pass
+            for test_report in test_reports:
+                full_text += test_report.longreprtext
```
gnikonorov, line 631: Can this ever be `None`? I ask since you add the following on line 278: `text = report.longreprtext or report.full_text`. Should we use `getattr` like we do below?

pytest_html/plugin.py
```diff
@@ -612,15 +615,57 @@ def _save_report(self, report_content):
...
+                    outcome = test_report.outcome
+                    failure_when = test_report.when
+
+                if hasattr(test_report, "wasxfail"):
+                    wasxfail = True
+
+                if test_report.outcome == "rerun":
+                    self.append_other(test_report)
+
+            test_report.outcome = outcome
```
gnikonorov, line 648: Should lines 648->665 be indented in one tab so that it's in the for loop?

gnikonorov (changes requested):
Overall this looks good, thank you @csm10495 and sorry for the very long delay here! I just had a few questions.

I'll be staying on top of this PR. If you haven't heard from me in a day or so, @ mention me so I see the notification

csm10495:
@gnikonorov, I made some fixes and added a comment. Tests pass feel free to merge if its good.

Though I will say that I wrote this a long time ago for a previous task, so I don't quite have the right mindset to fully test it again. Hoping the new unit tests (that i wrote back then) are good enough.

pytest_html/plugin.py
gnikonorov: even though this is valid, can you either add a comment stating this is intentional or use a different variable name? I thought it was a typo at first glance

pytest_html/plugin.py
```diff
@@ -618,15 +625,56 @@ def _save_report(self, report_content):
...
+            test_report.when = "call"
+            test_report.nodeid = test_name
+            test_report.longrepr = full_text
+            test_report.extra = extras
+            test_report.duration = duration
+
+            if wasxfail:
```
gnikonorov, line 665: can we simplify this to `test_report.wasxfail = wasxfail`?

Or can `test_report.wasxfail` still be `True` even if `wasxfail` is `False`?

pytest_html/plugin.py
```diff
@@ -618,15 +625,56 @@ def _save_report(self, report_content):
...
+                test_report.wasxfail = True
+
+            if test_report.outcome == "passed":
+                self.append_passed(test_report)
+            elif test_report.outcome == "skipped":
+                self.append_skipped(test_report)
+            elif test_report.outcome == "failed":
```
gnikonorov, line 672: Shouldn't we call `self.append_other(test_report)` as we were doing in the `else` case for `pytest_runtest_logreport`?

gnikonorov (changes requested):
Thanks for the prompt update, @csm10495! I had one more batch of comments.

Once those are addressed, I'd be happy to merge assuming @ssbarnea can take a look as well and OK this.

Also, can you add a changelog entry for this under `2.1.2 (unreleased)` in https://github.com/pytest-dev/pytest-html/blob/master/CHANGES.rst

pytest_html/plugin.py
csm10495: added comment for it

pytest_html/plugin.py
csm10495: It can't because other code relies on the attribute existing or not. So if we do the suggested, it would be false and exist.. and lead to more changes needed than desired to go into for this changeset.

gnikonorov: Thanks for clarifying @csm10495, I forgot about that

csm10495:
@gnikonorov , changes made. Hoping we're all good :)

gnikonorov (dismissed):
> @gnikonorov , changes made. Hoping we're all good :)

We are @csm10495, thank you for the prompt responses. I'll wait for @ssbarnea to take a look, but this LGTM to me

ssbarnea:
Please rebase. I do support the change, is very useful to have teardowns with results.

csm10495:
Thanks @ssbarnea and @gnikonorov. Squashed things down. Should be clean for merge now.

gnikonorov (approved):
LGTM, I'll leave the  merge to @ssbarnea so he can take a look at the final version too

ssbarnea (approved):

PR #278, opened 2020-03-05
Add download tag for self contained html

### LINKED ISSUE #273: Support FORMAT_TEXT on recent chrome with self contained html
Right now doing something like this:

```
import pytest

from pytest_html import extras

def test_ex(extra):
    print ("hey")
    extra.append(extras.text("Hello World!\n", "Hello Text"))
```

leads to something like this:

image

The problem is if you click the "Hello Text" link, chrome will close it immediately. This is a standard behavior as of recent chrome (https://stackoverflow.com/questions/46818033/opening-datatext-plain-window-in-chrome-showing-up-blank). 

Can we come up with a different way to make the text openable/viewable in self contained html?

csm10495:
This should resolve #273. If we're working with a self contained html, simply download text (and json) files instead of trying to open them in the current browser window.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -227,7 +227,7 @@ class HTMLReport:
                 content = json.dumps(extra.get("content"))
                 if self.self_contained:
                     href = data_uri(content, mime_type=extra.get("mime_type"))
-                    download = extra.get("name")
+                    download = extra.get("name") + '.json'
                 else:
                     href = self.create_asset(
                         content, extra_index, test_index, extra.get("extension")
@@ -239,7 +239,7 @@ class HTMLReport:
                     content = content.decode("utf-8")
                 if self.self_contained:
                     href = data_uri(content)
-                    download = extra.get("name")
+                    download = extra.get("name") + '.txt'
                 else:
                     href = self.create_asset(
                         content, extra_index, test_index, extra.get("extension")
@@ -252,13 +252,19 @@ class HTMLReport:
                 self._append_video(extra, extra_index, test_index)
 
             if href is not None:
+                kwargs = {
+                    "class_" : extra.get("format"),
+                    "href" : href,
+                    "target" : "_blank",
+                }
+
+                if download:
+                    kwargs['download'] = download
+
                 self.links_html.append(
                     html.a(
                         extra.get("name"),
-                        class_=extra.get("format"),
-                        href=href,
-                        target="_blank",
-                        download=download,
+                        **kwargs
                     )
                 )
                 self.links_html.append(" ")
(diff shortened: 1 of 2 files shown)


csm10495:
If we think the logic makes sense, I can work to try to fix the unit tests, otherwise, I'll leave this as is for now for further thoughts.

BeyondEvil:
This was to deal with something that Chrome (and IE/Edge) did (block data content).

I wonder if we only should change the behavior for those browsers (which would require us to deal with this in javascript) or if we sort of agree with the security aspect and change it for all the things?

I guess what I'm a little bit worried about is introducing a breaking change. 🤔

csm10495:
I think Firefox also blocks it...
https://blog.mozilla.org/security/2017/11/27/blocking-top-level-navigations-data-urls-firefox-59/

pytest_html/plugin.py
```diff
@@ -254,6 +258,7 @@ def append_extra_html(self, extra, extra_index, test_index):
                         class_=extra.get("format"),
                         href=href,
                         target="_blank",
+                        download=download,
```
christiansandberg: For non self-contained HTML this becomes `download="None"`. You probably need to add the argument conditionally somehow.

pytest_html/plugin.py
```diff
@@ -236,6 +239,7 @@ def append_extra_html(self, extra, extra_index, test_index):
                     content = content.decode("utf-8")
                 if self.self_contained:
                     href = data_uri(content)
+                    download = extra.get("name")
```
christiansandberg: You should add extension to the filename. Browsers seem to figure it out based on mime type but it is probably best to be explicit.

pytest_html/plugin.py
BeyondEvil:
```python
kwargs = {
    "class_": extra.get("format"),
    "href": href,
    "target": "_blank",
}

if download:
    kwargs["download"] = download

href = self.create_asset(
    content, extra_index, test_index, extra.get("extension"), **kwargs,
)
```

pytest_html/plugin.py
csm10495: Done (I think)

pytest_html/plugin.py
csm10495: Done (I think)

csm10495:
Added a couple tests and (hopefully) fixed the linting. @BeyondEvil give it one more look if you can. Thanks.

csm10495:
@ssbarnea any chance this or could get merged (after rebase)?

ssbarnea:
After rebase yes but it would also be very useful to provide a link in PR description to an sample file produced with this feature, so we can easily test how it works.

csm10495:
This fell wayyy off my radar. Closing out since i haven't been able to get to it. Sorry folks.

PR #281, opened 2020-03-11
Do not handle Tag elements as list in environment table

werdeil:
Use case: I want to pass a html.div element to the environment table.

With the current code, the element is seen as a list (as Tag inherits from list), and thus enter the list process which remove any html.* formatting.

Tell me if it is not the right solution.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,19 +1,41 @@
 Release Notes
 -------------
 
+**2.1.0 (2020-03-09)**
+
+* Added support for MP4 video format. (`#260 <https://github.com/pytest-dev/pytest-html/pull/260>`_)
+
+  * Thanks to `@ExaltedBagel <https://github.com/ExaltedBagel>`_ for the PR
+
+* Added support for sorting metadata by key. (`#245 <https://github.com/pytest-dev/pytest-html/issues/245>`_)
+
+  * Thanks to `@ssbarnea <https://github.com/ssbarnea>`_ for reporting and `@ExaltedBagel <https://github.com/ExaltedBagel>`_ for the fix
+
+* Added support for rendering reports collapsed (`#239 <https://github.com/pytest-dev/pytest-html/issues/239>`_)
+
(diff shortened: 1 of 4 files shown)


BeyondEvil:
Could you perhaps share a use-case with some examples? @werdeil

werdeil:
> Could you perhaps share a use-case with some examples? @werdeil

Of course: let say I want to display the plugins line differently.
Currently the line is {"html": "2.1.0", "metadata": "1.8.0"}
I can change the metadata through the metadata plugin's hook to have something prettier and have something like:

>  <table>
>   <tr>
>     <th>html</th>
>     <th>metadata</th>
>   </tr>
>   <tr>
>     <td>2.1.0</td>
>     <td>1.8.0</td>
>   </tr>
> </table>

which would be with the help of py.xml.html module:

>     html.table([
>         html.tr([
>             html.th("html"), 
>             html.th("metadata")
>         ]), 
>         html.tr([
>             html.td("2.1.0"), 
>             html.td("1.8.0")
>         ])
>     ])

Current implementation will lead to have the following line in the environment table, which is not I expected:

> "[<'tr' tag object 4460167632>, <'tr' tag object 4460169456>]"

BeyondEvil:
I wonder if the right thing to do here is to break out the [environment generation](https://github.com/pytest-dev/pytest-html/blob/master/pytest_html/plugin.py#L385) into its own class. Similar to [TestResult](https://github.com/pytest-dev/pytest-html/blob/master/pytest_html/plugin.py#L136).

That would make it easier to make modifications to the environment part of the report using hooks and fixtures. Future-proofing it basically.

A little more work, but better in the long run I think.

Would you be up for that?

werdeil:
I agree it would be better in the long run. I can try to do something but I'll need more time.

As you just released the 2.1.0, I guess I have the time to do it for next version.

werdeil:
Closing this PR, I'll push a new one with the environment generation class.

PR #285, opened 2020-03-20
Add divs for sections

werdeil:
Hi,

This PR add html.div element around each section to make them easier to manipulate through css or external javascript, for example hide or show some section.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -521,26 +521,36 @@ class HTMLReport:
 
         body = html.body(
             html.script(raw(main_js)),
-            html.h1(self.title),
-            html.p(
-                "Report generated on {} at {} by ".format(
-                    generated.strftime("%d-%b-%Y"), generated.strftime("%H:%M:%S")
+            html.div(
+                html.h1(self.title),
+                html.p(
+                    "Report generated on {} at {} by ".format(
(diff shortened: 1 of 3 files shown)


ssbarnea:
Makes sense, let me do some manual tests to assure it does not produce some undesired side-effects.

werdeil:
Thanks @ssbarnea for the review, I've tried to use summary and details tags but I find the result odd (see appended file (save as .txt as .html is not supported by Github)
[report.txt](https://github.com/pytest-dev/pytest-html/files/4573583/report.txt)

For my personal use it will need some css modification to have a correct visual but tell me what do you think and I'll push the changes to this PR.

On a side note I didn't find a way to have `<details open>` with py.xml, so I use `<details open="True">`instead, do you have another solution for this?

ssbarnea:
At this moment is is real PITA to test any PRs that modify the output because Travis does not have support for artifacts. I tried to use GitHub Actions and they only support zip files, which is again not friendly for us, as we want to host HTML.

I am inclined to enable AWS S3 upload to enable better testing.

werdeil:
@ssbarnea did you take a look at my appended report example, is it what you want with the `<details>` and `<summary>` parts?

BeyondEvil:
Closing this as I don't know how relevant it is for v4.

If you the reader feel like this is need and want to contribute a solution, please feel free to reopen.

PR #286, opened 2020-03-20
Environment table

werdeil:
Following remarks made on PR [281](https://github.com/pytest-dev/pytest-html/pull/281), here is a first version of the EnvironmentTable class which allow to use new hooks.

@BeyondEvil  tell me if it is what you had in mind.


pytest_html/plugin.py
```diff
@@ -551,30 +583,12 @@ def generate_summary_item(self):
...
-            elif isinstance(value, dict):
-                sorted_dict = {k: value[k] for k in sorted(value)}
-                value = json.dumps(sorted_dict)
-            raw_value_string = raw(str(value))
-            rows.append(html.tr(html.td(key), html.td(raw_value_string)))
-
-        environment.append(html.table(rows, id="environment"))
+        environment_table = self.EnvironmentTable(config).environment_table
+        if environment_table:
+            environment = [html.h2("Environment")]
```
BeyondEvil: This could be a title attribute on the EnvironmentTable class.

pytest_html/plugin.py
```diff
@@ -551,30 +583,12 @@ def generate_summary_item(self):
         return unicode_doc.decode("utf-8")
 
     def _generate_environment(self, config):
```
BeyondEvil: You can replace this entire function and where it is called just do: `self.EnvironmentTable(config).environment_table`

pytest_html/plugin.py
```diff
@@ -336,6 +336,38 @@ def _append_video(self, extra, extra_index, test_index):
             )
             self.additional_html.append(html.div(html_div, class_="video"))
 
+    class EnvironmentTable:
+        def __init__(self, config):
```
BeyondEvil: I would split this big init up into smaller functions.

Think builder pattern.

BeyondEvil:
Regarding builder pattern. We can have a default configuration and then users can use hooks to add to that config.

pytest_html/plugin.py
werdeil: Ok will be done in next version

pytest_html/plugin.py
werdeil: Yes will be done on the next version

pytest_html/plugin.py
werdeil: Not really sure to understand what to do here, I agree that everything in the init is not good but the function I can think of (generate_header, generate_row, ...) are very small... Do you want something like that?

BeyondEvil: Yeah, I'm having trouble articulating the vision I have in my head. Give me the weekend to see if I can throw something together that you can refine. :)

werdeil:
Hi @BeyondEvil  I just pushed something, tell me if it is what you are thinking when you say "builder pattern"

ssbarnea:
How about a rebase?

werdeil:
@ssbarnea Rebase done but there are failing tests, I guess they are new, I'll have a look

BeyondEvil:
I'd like to see a revamp of this for v4 if @werdeil is up for it.

PR #277, opened 2020-03-03
Fix image missing when using Base64 content

christiansandberg:
When adding a Base64 encoded image to a non-self contained report, only an empty link is created but no image.

PR #287, opened 2020-03-28
Add missing image/video tag

superdodd:
For image/video attachments that are not added as files, the non-self-contained html report does not show the image next to the test log as expected.  The div and link where the image/video goes is created, but no image or video element is inserted.


superdodd:
Whoops, this is a duplicate of #277 (and that one includes tests :))

PR #296, opened 2020-05-03
Use pytest-html when running its own tests

ssbarnea:
It does enable pytest-html reporting on its own test runs, so we
can easily see the outcome of the testing. This will also allow
publishing the results on CI/CD pipelines.

Reports are generated in the same location as tox logs, so they
do not overlap.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,7 +13,7 @@ deps =
     pytest-rerunfailures
     pytest-mock
     py{36,37,py3}-ansi2html: ansi2html
-commands = pytest -v -r a {posargs}
+commands = pytest -v -r a --color=yes --html={envlogdir}/report.html --self-contained-html {posargs}
 
 [testenv:linting]
 skip_install = True


ssbarnea:
@BeyondEvil ^ that would prove essential for testing outcome of incoming PRs. Bit of dogfooding would not hut.

BeyondEvil:
This is super nice @ssbarnea, thank you!

But before hitting the big green button on this, I want to figure out if it's possible to have Travis CI consume the report so that we can get the benefit there as well.

What do you think?

BeyondEvil:
Seems you had already done the leg work @ssbarnea !

https://docs.travis-ci.com/user/uploading-artifacts/

Thanks!

ssbarnea:
@BeyondEvil Travis has zero support for hosting or displaying any artifacts. This has being requested by users for years and I doubt them to ever do it.

The sad part is that others have similar issues, exception Zuul CI and Jenkins but node is viable for us.

I would prefer to find a quick solution for the moment because I am not sure how much time I have to invest in that. I was considering enabling upload of artifacts to one of my AWS S3 buckets (like s3.sbarnea.com), as this is very easy to implement in travis and I can afford to pay the hosting fees.

I did not propose it because I was almost sure someone would not like using my personal domain for something like this. 

Ideally we could make use of a `pytest.org` subdomain like `s3.pytest.org` or even better `logs.pytest.org` and feed build results there. I am wondering if we can persuade someone to help us doing this.

BeyondEvil:
I think we're talking passed each other.

I'm not talking about Travis directly supporting the storing (displaying) of artifacts.

I'm talking about the `artifacts addon`, which enables us to upload anything to S3 not being supported for PR test-runs.

Maybe I'm reading their docs wrong?

If I'm not, we could simply write our own "after script" that provides this for us.

ssbarnea:
@BeyondEvil I did not had time to investigate but lets merge this, at least helps testing outcomes locally.

PR #299, opened 2020-05-13
Better fix for TerminalReporter issue

BeyondEvil:
The [issue](https://github.com/pytest-dev/pytest/issues/6936) with TerminalWriter in `pytester` is causing tests to fail.

### DIFF
--- a/testing/test_pytest_html.py
+++ b/testing/test_pytest_html.py
@@ -14,20 +14,15 @@ import pytest
 pytest_plugins = ("pytester",)
 
 
-def handle_tr_writer_deprecation():
-    # Remove this function when they've fixed
+def remove_deprecation_from_recwarn(recwarn):
+    # TODO: Temporary hack until they fix
     # https://github.com/pytest-dev/pytest/issues/6936
-    import warnings
-    from _pytest.warnings import _setoption
-
-    arg = "ignore:TerminalReporter.writer:pytest.PytestDeprecationWarning"
-    _setoption(warnings, arg)
+    return [
+        item for item in recwarn if "TerminalReporter.writer" not in repr(item.message)
+    ]
 
 
 def run(testdir, path="report.html", *args):
-    # TODO: Temporary hack until they fix
-    # https://github.com/pytest-dev/pytest/issues/6936
-    handle_tr_writer_deprecation()  # TODO: Temporary hack
     path = testdir.tmpdir.join(path)
     result = testdir.runpytest("--html", path, *args)
     return result, read_html(path)
@@ -232,9 +227,6 @@ class TestHTML:
         assert report_title in html
 
     def test_report_title_addopts_env_var(self, testdir, monkeypatch):
-        # TODO: Temporary hack until they fix
-        # https://github.com/pytest-dev/pytest/issues/6936
-        handle_tr_writer_deprecation()
         report_location = "REPORT_LOCATION"
         report_name = "MuhReport"
         monkeypatch.setenv(report_location, report_name)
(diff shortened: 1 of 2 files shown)


testing/test_pytest_html.py
```diff
@@ -14,20 +14,15 @@
...
+    # TODO: Temporary hack until they fix
     # https://github.com/pytest-dev/pytest/issues/6936
-    import warnings
-    from _pytest.warnings import _setoption
-
-    arg = "ignore:TerminalReporter.writer:pytest.PytestDeprecationWarning"
-    _setoption(warnings, arg)
+    for index, item in enumerate(recwarn):
+        if "TerminalReporter.writer" in repr(item.message):
+            del recwarn.list[index]
```
dhalperi, line 22: I don't think this is really safe if there are multiple instances in the list, but in practice that likely does not matter here.

(if you delete index 0 in a list of length 2, enumerate will never even product the second item).

BeyondEvil: Aha, is it better to store the relevant indexes and then after the iteration is complete delete them? @dhalperi

dhalperi, line 22: I might do something like this: (sorry for laziness, couldn't get github multiline suggest to work)

```py
def filter_warnings(recwarn):
  # TODO..
  return [w for w in recwarn.list if "TerminalReporter.writer" in repr(w.message)]
```

and then below

```py
warnings = filter_warnings(recwarn)
assert len(warnings) == 0
```

dhalperi: see suggestion - basically, for me, not modifying recwarn seems easier than tracking indices, etc.

BeyondEvil: Thanks! 🙏

BeyondEvil:
Seems like Travis is having some issues. Will try again tomorrow.

BeyondEvil:
@nicoddemus Sorry for spamming with reviewers, but this is blocking other PRs.

testing/test_pytest_html.py
```diff
@@ -894,9 +887,6 @@ def test_css_invalid(self, testdir, recwarn):
         assert "No such file or directory: 'style.css'" in result.stderr.str()
 
     def test_css_invalid_no_html(self, testdir):
```
nicoddemus, line 889: If you mark the tests to filter warnings:

```python
@pytest.mark.filterwarnings("ignore:.*TerminalReporter.writer.*")
```

You might not need `remove_deprecation_from_recwarn` in the end?

nicoddemus:
> @nicoddemus Sorry for spamming with reviewers, but this is blocking other PRs.

No worries, I appreciate the ping. 👍

testing/test_pytest_html.py
BeyondEvil: Tried it and it didn't work @nicoddemus 

Maybe I'm doing something wrong?

```python
    @pytest.mark.filterwarnings("ignore:.*TerminalReporter.writer.*")
    @pytest.mark.parametrize("colors", [(["red"]), (["green", "blue"])])
    def test_css(self, testdir, recwarn, colors):
        css = {}
        cssargs = []
        for color in colors:
            style = f"* {{color: {color}}}"
            path = testdir.makefile(".css", **{color: style})
            css[color] = {"style": style, "path": path}
            cssargs.extend(["--css", path])
        result, html = run(testdir, "report.html", "--self-contained-html", *cssargs)
        assert result.ret == 0
        assert len(recwarn) == 0
        for k, v in css.items():
            assert str(v["path"]) in html
            assert v["style"] in html
```

nicoddemus: Yeah unfortunately the mark is considered only for the test run, not the setup (which is where the warning is issued).

PR #311, opened 2020-06-27
Customizable html output structure

memetb:
Following up on #310, I've made a basic modification to the html structure and how it is generated.

I also added a new hook called `pytest_html_report_final_dom`, the semantics of which are fairly straightforward. 

Commit e8bd795 in itself would be sufficient for addressesing issue #310. Commit 8c97045 is an added bonus to make life easier.

I would also recommend that perhaps in the future, a migration away from `py.xml.html` might be beneficial (in favor of core python libraries that support native tree walking etc). The [py](https://github.com/pytest-dev/py) package is noted as being in maintenance mode and new projects are recommended not to be started using that library. 

Given that the changes required to upgrade `pytest-html` are actually somewhat limited, it might be a good migration target for the future.

Please let me know if or how I should amend this PR to meet project guidelines.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -259,6 +259,27 @@ additional HTML and log output with a notice that the log is empty:
           del data[:]
           data.append(html.div('No log output captured.', class_='empty log'))
 
+Modifying the whole test document
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Finally, for major customizations, you may manipulate and update the resulting
+html document prior to finalization by hooking :code:`pytest_html_report_final_dom`.
+
+.. code-block:: python
+
+  import pytest
+
+  def pytest_html_report_final_dom(document):
+      from importlib.resources import read_text
+      js = read_text(".", "fancy.js")
+      for node in list(iter(document)):
+          if isinstance(node, html.head):
+             node.extend([html.script(raw(js), type="text/javascript")])
+
+
+
+
+
 Display options
 ---------------
 
--- a/README.rst
+++ b/README.rst
@@ -269,7 +269,8 @@ html document prior to finalization by hooking :code:`pytest_html_report_final_d
 
   import pytest
 
-  def pytest_html_report_final_dom(document):
+  def pytest_html_report_final_dom(session, document):
+      # session is a `_pytest.main.Session` object
       from importlib.resources import read_text
       js = read_text(".", "fancy.js")
       for node in list(iter(document)):
--- a/pytest_html/hooks.py
+++ b/pytest_html/hooks.py
@@ -2,8 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from py.xml import html
-
 
 def pytest_html_report_title(report):
     """ Called before adding the title to the report """
@@ -25,5 +23,5 @@ def pytest_html_results_table_html(report, data):
     """ Called after building results table additional HTML. """
 
 
-def pytest_html_report_final_dom(document: html.html):
+def pytest_html_report_final_dom(document):
     """ Called after the entire document has been created. """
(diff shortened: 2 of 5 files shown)


memetb:
Sorry, I can see I didn't follow the instructions in dev. I'll do that shortly. Feel free to comment in the meanwhile.

memetb:
I might need some help with the npm/grunt tests. 

Also, I'm seeing 45 tests failing on `master`.  

Comments?

BeyondEvil:
This is really nice work @memetb ! 💪 

I would really like @ssbarnea input on this as well. 😊 

> I would also recommend that perhaps in the future, a migration away from py.xml.html might be beneficial (in favor of core python libraries that support native tree walking etc). The py package is noted as being in maintenance mode and new projects are recommended not to be started using that library.

Yes, I agree 100% and it's been on my ToDo for a long time - I just have not had the time. :(

Are there core python libraries that support HTML natively?

memetb:
Thanks. Much appreciated.

With regards to html libraries: I really like the whole `html.div( blah blah )` verbosity as a library *consumer*, i.e. when I just want to quickly spit out some html. However, as a library provider, I think the most important thing should be walkability/visitability.

So very high up on my list of requirements for that `document` variable that's passed to the hook is that I could use an xpath query on it. 

To that effect, the [xml](https://docs.python.org/3/library/xml.etree.elementtree.html#parsing-xml) module that comes stock with python is pretty powerful.

I can take that `document` object as it currently is and simply do:

    import xml.etree.ElementTree as ET 
    doc = ET.fromstring(document.unicode())
    doc.findall(".//div") # xpath query

To output from an element tree, one would have to do an `io.ByteBuffer` to get a string, or can simply call:

    ET.ElementTree(doc).write( filename )

The weakness of `ElementTree` comes from the building of the tree (it's not as neat and easy as `py.xml`), but I think that's a bullet we can bite as a library provider.

BeyondEvil:
> Thanks. Much appreciated.
> 
> With regards to html libraries: I really like the whole `html.div( blah blah )` verbosity as a library _consumer_, i.e. when I just want to quickly spit out some html. However, as a library provider, I think the most important thing should be walkability/visitability.
> 
> So very high up on my list of requirements for that `document` variable that's passed to the hook is that I could use an xpath query on it.
> 
> To that effect, the [xml](https://docs.python.org/3/library/xml.etree.elementtree.html#parsing-xml) module that comes stock with python is pretty powerful.
> 
> I can take that `document` object as it currently is and simply do:
> 
> ```
> import xml.etree.ElementTree as ET 
> doc = ET.fromstring(document.unicode())
> doc.findall(".//div") # xpath query
> ```
> 
> To output from an element tree, one would have to do an `io.ByteBuffer` to get a string, or can simply call:
> 
> ```
> ET.ElementTree(doc).write( filename )
> ```
> 
> The weakness of `ElementTree` comes from the building of the tree (it's not as neat and easy as `py.xml`), but I think that's a bullet we can bite as a library provider.

Interesting! 🤔

memetb:
@BeyondEvil: I've been dogfooding this addition, and I think the hook should for the dom finalizer should include a session parameter to avoid ugly globals. I will add a final commit to this affect.

memetb:
Any notes or comments on this PR?

ssbarnea:
@memetb Please resolve conflicts.

memetb:
@ssbarnea: Merge is done, however I'll have to check what has changed since I wrote this. evidently something no longer passes the node test.

memetb:
Fellas, I'm at a loss for what's going wrong here. 

The `show_hide_extras` test is opaque to me in what it's trying to do, and it isn't my code. I actually don't see how it's even related to the code in this branch given that it [finds the element](https://github.com/memetb/pytest-html/blob/647616c3fc51b9a671c9427d835c6f558b049060/testing/test.js#L97) based on class name ('.passed').

Likewise, the `sort_column_test` error I'm seeing is seemingly unrelated to this revision.

A second pair of eyes would be appreciated here.

BeyondEvil:
As the plugin as seen major rework in v4, I'm not sure how relevant this is anymore.

If you the reader want to contribute with something here, please feel free to reopen.

PR #313, opened 2020-07-07
Make the maximum asset filename length configurable.

D3X:
Some filesystems do not support filenames that are 255 characters long. Notably, [aufs](https://en.wikipedia.org/wiki/Aufs) used by CircleCI only supports filenames up to 242 characters.
By making this a configurable option, we allow users to set their desired length limit.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -69,6 +69,12 @@ def pytest_addoption(parser):
         default=False,
         help="Open the report with all rows collapsed. Useful for very large reports",
     )
+    parser.addini(
+        "max_asset_filename_length",
+        default=255,
+        help="set the maximum filename length for assets "
+        "attached to the html report.",
+    )
 
 
 def pytest_configure(config):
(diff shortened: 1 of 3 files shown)


BeyondEvil:
Awesome!

Should this be an ini-option instead of a command line option? 🤔

D3X:
@BeyondEvil I have changed it to be an ini option and force-pushed. It's failing linting, because `flake8` has a different rule than `black` regarding whitespace before colons. Probably something to be fixed in a separate PR.

BeyondEvil:
> @BeyondEvil I have changed it to be an ini option and force-pushed. It's failing linting, because `flake8` has a different rule than `black` regarding whitespace before colons. Probably something to be fixed in a separate PR.

Interesting, I'll pull your PR and check it out.

BeyondEvil:
@D3X 

Please just add `ignore = E203` under the `[flake8]` section in `tox.ini`

BeyondEvil:
Sorry @D3X I should've been more clear, apart from adding `ignore = E203` you also have to put that space before colon. That's why linting failed now.

As a side note, are you using `pre-commit` with this project? It would've caught the above before committing. Instructions [here](https://github.com/pytest-dev/pytest-html/blob/master/development.rst#development)

D3X:
@BeyondEvil sorry it took so long, I was quite busy the last couple of weeks. The build's green now, so hopefully we can get this merged and released soon-ish ;)

Issue #314, opened 2020-07-13
Garbage date-time is printed in the Captured log.


pawan7476:
Garbage date-time is printed in the Captured log.

**pytest-html report:**

------------------------------ Captured log setup ------------------------------
[32mINFO    [0m root:test_cyclic_switchover.py:26 Inside Setup 
[32mINFO    [0m root:test_cyclic_switchover.py:54 
------------------------------ Captured log call -------------------------------
[32mINFO    [0m root:test_cyclic_switchover.py:82 
 Switchover Process : SCM

**pytest console log :** 

collected 4 items / 2 deselected / 2 selected

test_cyclic_switchover.py::test_process_switchover[SCM]
------------------------------------------------------------ live log setup -------------------------------------------------------------
2020-07-13 21:51:50 [    INFO] Inside Setup
 (test_cyclic_switchover.py:26)
2020-07-13 21:51:50 [    INFO]
(test_cyclic_switchover.py:54)
------------------------------------------------------------- live log call -------------------------------------------------------------
2020-07-13 21:51:50 [    INFO]
 Switchover Process : SCM (test_cyclic_switchover.py:82)
PASSED                                                                                                                            [ 50%]
test_cyclic_switchover.py::test_process_switchover[SAM]
------------------------------------------------------------- live log call -------------------------------------------------------------
2020-07-13 21:51:50 [    INFO]
 Switchover Process : SAM (test_cyclic_switchover.py:82)
PASSED                                                                                                                            [100%]
----------------------------------------------------------- live log teardown -----------------------------------------------------------
2020-07-13 21:51:50 [    INFO] Inside Teardown (test_cyclic_switchover.py:60)

BeyondEvil:
Could you provide some more info @pawan7476 ?

- OS and version
- Python version
- Pytest version
- Plugin version
- Default encoding of your OS

It'd be great if you could also provide a minimum reproducible test.

Thanks!

pawan7476:
Hi @BeyondEvil,

Please find the required details : 

- py==1.9.0
- pycparser==2.20
- PyNaCl==1.4.0
- pyparsing==2.4.7
- pytest==5.4.3
- pytest-html==1.22.1
- pytest-metadata==1.8.0
- pytest-repeat==0.8.0
- Python | 3.5.2
- Packages | {'pluggy': '0.13.1', 'pytest': '5.4.3', 'py': '1.9.0'}
[...]

PR #315, opened 2020-07-16
Strip ANSI escape sequences when ansi2html is missing

BeyondEvil:
Fixes: #314

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,6 +24,9 @@ commands = pre-commit run --all-files --show-diff-on-failure
 [flake8]
 max-line-length = 88
 exclude = .eggs,.tox
+# rational here:
+# https://github.com/psf/black/blob/master/docs/the_black_code_style.md#slices
+extend-ignore = E203
 
 [pytest]
 testpaths = testing
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,7 @@ commands = pre-commit run --all-files --show-diff-on-failure
 [flake8]
 max-line-length = 88
 exclude = .eggs,.tox
-# rational here:
+# rationale here:
 # https://github.com/psf/black/blob/master/docs/the_black_code_style.md#slices
 extend-ignore = E203
 
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -30,6 +30,8 @@ from py.xml import html, raw
 from . import extras
 from . import __version__, __pypi_url__
 
+from _pytest.logging import _remove_ansi_escape_sequences
+
 
 def pytest_addhooks(pluginmanager):
     from . import hooks
@@ -279,9 +281,13 @@ class HTMLReport:
                 header, content = map(escape, section)
                 log.append(f" {header:-^80} ")
                 log.append(html.br())
+
                 if ANSI:
                     converter = Ansi2HTMLConverter(inline=False, escaped=False)
                     content = converter.convert(content, full=False)
+                else:
+                    content = _remove_ansi_escape_sequences(content)
+
                 log.append(raw(content))
                 log.append(html.br())
 
--- a/testing/test_pytest_html.py
+++ b/testing/test_pytest_html.py
@@ -830,6 +830,29 @@ class TestHTML:
             else:
                 assert content not in html
 
+    def test_ansi_escape_sequence_removed(self, testdir):
+        testdir.makeini(
+            r"""
+            [pytest]
+            log_cli = 1
+            log_cli_level = INFO
+        """
+        )
+        testdir.makepyfile(
+            r"""
+            import logging
+            logging.basicConfig()
+            LOGGER = logging.getLogger()
+            def test_ansi():
+                LOGGER.info("ANSI removed")
+        """
+        )
+        result, html = run(
+            testdir, "report.html", "--self-contained-html", "--color=yes"
+        )
+        assert result.ret == 0
+        assert not re.search(r"\[[\d;]+m", html)
+
     @pytest.mark.parametrize("content", [("'foo'"), ("u'\u0081'")])
     def test_utf8_longrepr(self, testdir, content):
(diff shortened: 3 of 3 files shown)


tox.ini
```diff
@@ -22,6 +22,7 @@ deps = pre-commit
 commands = pre-commit run --all-files --show-diff-on-failure
 
 [flake8]
+ignore = E203
```
ssbarnea, line 25: I see not reason for this, also no explanation. If is a bug in flake8, why not using an inline exclude instead of diabling an entire rule?

BeyondEvil:
> I have mixed feelings regarding this because it goes a dangerous path (auto-processing input, not even mentioning why is stipping ANSI.)
> 
> Maybe we should log that we remove ANSI as ansi2html is not present.

Why do you figure? To me this is an obvious bug, no? I think emitting a warning in this case would just add noise.

> Adding E203 as general exception is not acceptable. Inline exclude would be ok if not avoidable other way and if added as general exception, it must be preceded with a comment line explaining the reasoning for doing it. The only good reason I remember was incompatibility with black.

Apparently E203 is [broken](https://github.com/psf/black/issues/565#issuecomment-429877688) and not PEP8 compliant, so I don't see any reason to not disable it completely. Especially since `black` seems to handle it correctly.
 
> Also please rephrase it as "Strip ANSI when ansi2html is missing", it would be easier to understand why this needs to happen.

You mean rephrase the PR?

tox.ini
BeyondEvil: Afaik, the rule is broken full stop, not in particular cases. It's also not PEP8 compliant.

ssbarnea:
> You mean rephrase the PR?

Yeah. Usually I would have rephrased the PR myself but I do not have this privilege. As these end-up as release notes, I do not want to force user to raise a brow when reading it.

Regarding printing a warning, is was only a suggestion, not a request. It is very common for new users of pytest-html to miss having ansi2html, as is not an implicit requirement. Printing something, could make them fix the issue instead of wondering around.

BeyondEvil:
> > You mean rephrase the PR?
> 
> Yeah. Usually I would have rephrased the PR myself but I do not have this privilege.

No problem, I'll rephrase.

> As these end-up as release notes, I do not want to force user to raise a brow when reading it.

Not sure what you mean, but there's no automation surrounding the release notes (yet). It's all manually written.

> Regarding printing a warning, is was only a suggestion, not a request. It is very common for new users of pytest-html to miss having ansi2html, as is not an implicit requirement. Printing something, could make them fix the issue instead of wondering around.

Oh, I see. 🤔 Do you mean that the default usage of the plugin should be to use it in conjunction with `ansi2html`?

If that's the case, we could emit a more general warning saying that it's recommended to use it together with the plugin. @ssbarnea

tox.ini
```diff
@@ -24,6 +24,9 @@ commands = pre-commit run --all-files --show-diff-on-failure
 [flake8]
 max-line-length = 88
 exclude = .eggs,.tox
+# rationale here:
+# https://github.com/psf/black/blob/master/docs/the_black_code_style.md#slices
+extend-ignore = E203
```
BeyondEvil, line 29: ping @ssbarnea

ssbarnea: I already approved but apparently I do not get enough rights for the vote to matter (is gray, not green).

nicoddemus:
>  Do you mean that the default usage of the plugin should be to use it in conjunction with ansi2html?

How about making `ansi2html` an explicit requirement?

BeyondEvil:
> > Do you mean that the default usage of the plugin should be to use it in conjunction with ansi2html?
> 
> How about making `ansi2html` an explicit requirement?

Has something changed re. licensing?

https://github.com/pytest-dev/pytest-html/pull/97

@nicoddemus

nicoddemus:
Oh forgot about that issue, thanks for reminding me.

nicoddemus:
A quick search turns up https://bitbucket.org/dhrrgn/ansiconv, but I'm can't vouch if it is maintained.

BeyondEvil:
> A quick search turns up https://bitbucket.org/dhrrgn/ansiconv, but I'm can't vouch if it is maintained.

Hasn't been touched since 2014.

Do you see an issue with my solution? It's straight from `pytest` itself.

nicoddemus:
Not in particular other than using a private function, otherwise LGTM!

ssbarnea:
TO be honest, I had an years old with to rewrite ansi2html with a MIT/BSD-like license that does not impose reusability limitations but never had time. The amount of code it has is very low. Still, I always have other more pressing issues to sort, maybe if we join efforts we can make it a reality?

I even bootstrapped it at https://github.com/pycontribs/a2h -- pypi package namespace too. With your help we can probably sort it quickly. There is a long list of python libraries that share the same issue around ansi2html, is not only pytest-html that would benefit from it.

BeyondEvil:
> TO be honest, I had an years old with to rewrite ansi2html with a MIT/BSD-like license that does not impose reusability limitations but never had time. The amount of code it has is very low. Still, I always have other more pressing issues to sort, maybe if we join efforts we can make it a reality?
> 
> I even bootstrapped it at https://github.com/pycontribs/a2h -- pypi package namespace too. With your help we can probably sort it quickly. There is a long list of python libraries that share the same issue around ansi2html, is not only pytest-html that would benefit from it.

I'm happy to help. But you would have to take lead.

Also, this plugin needs a rewrite as well. Especially moving away from the py.xml dependency.

ssbarnea:
I sent invites to both of you. I will take case of pushing the code and setting up CI/CD pipelines for it. If I am lucky I may have something ready before Monday.

ssbarnea:
Sadly the plan to work on a2h over the weekend had to be postponed as I faced other more pressing fixing, like creating a maintenance fork for a ruamel.yaml library. Still, I will try to do something during next week.

PR #319, opened 2020-07-29
Avoid pytest 6.0.0

ssbarnea:
Avoid incompatibility issues caused by https://github.com/pytest-dev/pytest/issues/7559

We only avoid 6.0.0 because a fix will be included in 6.0.1


ssbarnea:
The reason for unexpected branch name is that I had other plans for it, but those were delayed until we got tests running.

ssbarnea:
@BeyondEvil please merge, needed to bring CI to green and to test the other changes.

PR #320, opened 2020-07-29
Enable devel testing job

ssbarnea:
Start testing pytest with code from pytest master and with prereleases of other dependencies, allowing us to detect breakages before upstream releases are made.

This change needs https://github.com/pytest-dev/pytest-html/pull/319 merged first in order to get CI/CD green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,12 @@ jobs:
       env: TOXENV=py37-ansi2html
 
     -
+      python: 3.8
+      dist: xenial
+      sudo: required
+      env: TOXENV=devel
+
+    -
       python: pypy3
       env: TOXENV=pypy3
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: python
 jobs:
+  allow_failures:
+    - name: devel
+      python: 3.8
+      dist: xenial
+      env: TOXENV=devel
+
   include:
     - stage: tests
       language: node_js
@@ -37,10 +43,9 @@ jobs:
       sudo: required
       env: TOXENV=py37-ansi2html
 
-    -
+    - name: devel
       python: 3.8
       dist: xenial
-      sudo: required
       env: TOXENV=devel
 
     -
--- a/tox.ini
+++ b/tox.ini
@@ -21,6 +21,14 @@ basepython = python3
 deps = pre-commit
 commands = pre-commit run --all-files --show-diff-on-failure
 
+[testenv:devel]
+description = Tests with unreleased deps
+basepython = python3
+pip_pre = True
+deps =
+    {[testenv]deps}
+    pytest @ git+https://github.com/pytest-dev/pytest.git
+
 [flake8]
 max-line-length = 88
 exclude = .eggs,.tox


.travis.yml
```diff
@@ -37,6 +37,12 @@ jobs:
       sudo: required
       env: TOXENV=py37-ansi2html
 
+    -
+      python: 3.8
+      dist: xenial
+      sudo: required
```
BeyondEvil, line 43: `sudo` is apparently default now (at least for `xenial`), so you can remove this line.

.travis.yml
```diff
@@ -37,6 +37,12 @@ jobs:
...
+      env: TOXENV=devel
+
```
BeyondEvil, line 45: I would like for this job to only run on `master` branch.

```
branches:
  only:
  - master
```

We've setup a cron-job to run daily.

I don't want this job to run on PRs, because I don't think that burden (if it fails) should land on the PR-creator. Especially since a failure can be unrelated to the changes in the PR.

That's for us maintainers to deal with.

ssbarnea: @BeyondEvil your personal preference makes it impossible to validate that this change does anything. I do not trust myself to merge stuff in the hope they will run correctly later.

Still, I updated travis config to allow failure of devel.

BeyondEvil: Not sure what that means... 🤔

Why was running this only on master an issue?

I expect that when it fails, people (admins, members, etc.) are notified?

However, that raises an interesting question. What I would like to see is that if the CRON job fails, an issue is automatically created.

Moving to GA should definitely solve that, yes?

I’m sure such an integration ex

BeyondEvil: Oh, I think I see what you’re saying (I would’ve worded it differently tho).

No worries, I have a PR in the works to add and update to python 3.8 everywhere. I can do it in that.

ssbarnea: Nope, cron execution does not create tickets neither on travis or github-actions. In theory it can be implemented but it has some security risks as it requires tokens with write access to github. Also there are some other practical reasons why I never seen it implemented: it is too easy to get the repo spammed with broken build reports.

That is why everyone I seen relies on email send from CI on failure. If I am correct they do email cores by default.

Regarding "running only on master" question, let me explain. Look at travis checks on current change, you do see "devel" among them, right? Clearly we do want to see that "tox -e devel" runs on current change? If we configure it to run only on master branch, we would not be able to see the result of devel run here anymore, making impossible to test any fix related to devel, as we would be driving blindly, we would know the result only after the merge.

.travis.yml
ssbarnea: done

ssbarnea:
Note: if you do not want py38, we can use something else, I only picked it in order to increase the testing coverage a bit, no strong feelings about which version is used. Some hardcore testers would likely put a 3.9 pre-release there ;)

BeyondEvil:
> Note: if you do not want py38, we can use something else, I only picked it in order to increase the testing coverage a bit, no strong feelings about which version is used.

No no, py38 is great!

> Some hardcore testers would likely put a 3.9 pre-release there ;)

We should actually have that _as well_, but not in the same test. The reason being that if something fails we'll first have to figure out if it's due to pytest or python, having them separate removes that step.

.travis.yml
BeyondEvil: > Nope, cron execution does not create tickets neither on travis or github-actions.

I know that. What I meant was that someone might have created an integration you could install/activate.

> In theory it can be implemented but it has some security risks as it requires tokens with write access to github. 

Well sure, but the GH tokens can be tightly scoped. Already today, the world can create an issue in the project.

> Also there are some other practical reasons why I never seen it implemented: it is too easy to get the repo spammed with broken build reports.

Well sure, but that spamming occurs in email-boxes today.

PR #326, opened 2020-08-10
Enable py38 testing

ssbarnea:
- Advertise support for py38
- Adds CI/CD jobs for testing py38

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,20 @@ jobs:
       env: TOXENV=py37
 
     -
+      python: 3.8
+      dist: xenial
+      env: TOXENV=py38
+
+    -
       python: 3.7
       dist: xenial
       sudo: required
       env: TOXENV=py37-ansi2html
+
+    -
+      python: 3.8
+      dist: xenial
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py{36,37,py3}{,-ansi2html}, linting
+envlist = py{36,37,38,py3}{,-ansi2html}, linting
```
hugovk, line 7: Also add to line 15

.travis.yml
```diff
@@ -31,12 +31,22 @@ jobs:
...
+
     -
       python: 3.7
       dist: xenial
       sudo: required
       env: TOXENV=py37-ansi2html
 
+    -
+      python: 3.8
+      dist: xenial
```
hugovk, line 47: `dist: xenial` is the default, so could be removed from all of these.

Also `sudo:` no longer does anything and can be removed: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Issue #327, opened 2020-08-10
proposal: combine ansi2html and non-ansi2html tox envs into one


ssbarnea:
We should be able to test pytest-html behavior in presence or absence of ansi2html without having two different tox environments.

By avoiding a new dimention in tox env matrix, we can simplify local and CI/CD testing, making much easier to maintain the tool. Absence of ansi2html cab easily be mocked.

Please add +/- to it and based on that I can file a PR, or not.

BeyondEvil:
I'm interested to see a PR on this.

PR #328, opened 2020-08-11
Consolidate tox environments

ssbarnea:
Simplifies testing matrix by mocking the absence of ansi2html support.

Fixes: #327

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -4,6 +4,8 @@
 
 from base64 import b64encode, b64decode
 from collections import OrderedDict
+from functools import lru_cache
+import importlib
 from os.path import isfile
 import datetime
 import json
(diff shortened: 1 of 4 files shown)


pytest_html/plugin.py
```diff
@@ -33,6 +27,16 @@
...
 
 
+@lru_cache()
+def ansi_support():
+    try:
+        # from ansi2html import Ansi2HTMLConverter, style  # NOQA
+        return importlib.import_module("ansi2html")
+    except ImportError:
+        # ansi2html is not installed
+        pass
```
BeyondEvil, line 37: Looking at the usage of this function, could we have an explicit return here? Either `False` or `None`?

testing/test_pytest_html.py
```diff
@@ -809,14 +809,14 @@ def test_foo(val):
...
-        except ImportError:
-            # ansi2html is not installed
-            ANSI = False
+    @pytest.mark.parametrize(
+        "with_ansi", [True, False],
+    )
+    def test_ansi_color(self, testdir, mocker, with_ansi):
+        if not with_ansi:
+            mock_ansi_support = mocker.patch("pytest_html.plugin.ansi_support")
+            mock_ansi_support.return_value = None
```
BeyondEvil, line 818: I believe you can do: 

```python
if not with_ansi:
    mocker.patch("pytest_html.plugin.ansi_support", return_value=None)
```

pytest_html/plugin.py
```diff
@@ -33,6 +27,16 @@
 from _pytest.logging import _remove_ansi_escape_sequences
 
 
+@lru_cache()
```
BeyondEvil, line 30: Nice!

pytest_html/plugin.py
ssbarnea: Possible but not really needed. The implicit return None works really nice.

PR #330, opened 2020-08-11
Enable github actions

ssbarnea:


ssbarnea:
@BeyondEvil after merging remember to edit settings of master branch and add protection to require all github actions to be green (now only travis is "required" to pass)

.github/workflows/tox.yml
```diff
@@ -0,0 +1,43 @@
+name: gh
+
+on:
+  create:  # is used for publishing to PyPI and TestPyPI
```
nicoddemus, line 4: I don't see any deployment being done here, are those comments meant for later or am I missing something?

BeyondEvil: I think the idea is for this to run along side Travis for a while, so we'll let Travis to the deploy until we switch over. @nicoddemus 

Correct me if I'm wrong @ssbarnea

ssbarnea: Yes. That is the plan. In fact code for deploying using github actions is missing from current change. We can switch to github publishing only when we are sure we want to drop travis.

nicoddemus: OK, thanks folks.

PR #335, opened 2020-09-17
Sync .pre-commit-config.yaml with pytest

gnikonorov:
Sync the ` .pre-commit-config.yaml` file with `pytest`'s as much as possible for a better developer experience.

The following diffs that cause `pre-commit` behavior differences between the two `pre-commit` workflows still exist:
```
gnikonorov:~/pytest-html$ diff .pre-commit-config.yaml ../pytest/.pre-commit-config.yaml
...
>           - flake8-docstrings==1.5.0
...
> -   repo: https://github.com/asottile/setup-cfg-fmt
>      ...
> -   repo: https://github.com/pre-commit/mirrors-mypy
        ...
46a61,82
>     -   id: changelogs-rst
>          ....
>     -   id: py-deprecated
>         ...
gnikonorov:~/pytest-html$
```

This is why:
- `flake8-docstrings`: requires a big documentation push, so I deferred
- `setup-cfg-fmt`: requires changes to `setup.cfg` which I believe should tackled separately
- `mirrors-mypy`: requires typing. I would like to add this, but I think it is beyond the scope of this PR
- `changelogs-rst`: ends up skipping the changelog is not compatible with this repos changelog format
- `py-deprecated`: is not applicable to this project

@BeyondEvil @ssbarnea if neither of you object, I'll raise an issue to add typing/mypy + any of the other above targets we think are useful

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,6 @@
-# NOTE: We should still add setup-cfg-fmt and mypy, but they require extra work to implement
 repos:
 -   repo: https://github.com/psf/black
-    rev: stable
+    rev: 19.10b0
     hooks:
     -   id: black
         args: [--safe, --quiet]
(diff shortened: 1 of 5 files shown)


ssbarnea (approved):

ssbarnea:
@gnikonorov Extra kudos for splitting in chunks!

BeyondEvil:
Thanks @gnikonorov, you're a rock star! ⭐

PR #337, opened 2020-09-19
Enable/fix Windows testing, enable Mac/Windows in Github Actions: Closes #336

### LINKED ISSUE #336: testing fails on Windows
Running unit tests fails on Windows:

testing/test_pytest_html.py::TestHTML::test_utf8_longrepr[u'\x81'] FAILED             
and
testing/test_pytest_html.py::TestHTML::test_report_display_utf8 FAILED

I wonder if we can/should skip these tests on Windows or modify them in some way.

csm10495:
Fix the default encoding by patching open to always use utf-8. This is the default on *nix based systems, though is different on Windows (and possibly others). 

Also while here, add testing for Mac/Windows to Github Actions

### DIFF
--- a/pytest_html/extras.py
+++ b/pytest_html/extras.py
@@ -10,10 +10,10 @@ FORMAT_URL = "url"
 FORMAT_VIDEO = "video"
 
 
-def extra(content, format, name=None, mime_type=None, extension=None):
+def extra(content, format_type, name=None, mime_type=None, extension=None):
     return {
         "name": name,
-        "format": format,
+        "format_type": format_type,
         "content": content,
         "mime_type": mime_type,
         "extension": extension,
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -15,21 +15,65 @@ on:
 jobs:
   build:
     name: ${{ matrix.name }}
-    runs-on: ubuntu-18.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         include:
-          - name: linting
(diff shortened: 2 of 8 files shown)


csm10495:
@ssbarnea can you check this?

ssbarnea (changes requested):
You must use different names for jobs in order to avoid confusions, example: "py3x-(windows|macos|linux)".

csm10495:
@ssbarnea I finally got around to fixing this. Check it again. Right now it fails because of https://github.com/actions/setup-python/issues/151 (I don't think its my change specifically). 

Also in order to merge I think you may need to change settings to change the required actions for merge (since I changed the names):
image

gnikonorov:
Hey @csm10495 can you resolve merge conflicts, then I can take  a look

csm10495:
Yep working on it and the commit rewrite.

pytest_html/extras.py
```diff
@@ -10,10 +10,10 @@
 FORMAT_VIDEO = "video"
 
 
-def extra(content, format, name=None, mime_type=None, extension=None):
```
gnikonorov, line 13: I think the rebase is a little off. This change already exists in master ( see https://github.com/pytest-dev/pytest-html/blob/master/pytest_html/extras.py#L13 ). It looks like this PR counts the merge of master as your changes

gnikonorov (changes requested):

csm10495:
Closing in favor of a new PR.

PR #347, opened 2020-10-17
Stop shadowing the 'format' builtin

gnikonorov:
Stop shadowing the `format` builtin function, since this is bad practice.


ssbarnea (approved):
Ok but I labeled it as major due to API change, so we will be reminded to bump major versions when we release. I doubt anyone is really using this plugin API but is better to make it clear and to avoid breaking semantic versioning users.

PR #349, opened 2020-10-17
Support tox isolated builds

gnikonorov:
Support [isolated builds](https://tox.readthedocs.io/en/latest/config.html#conf-isolated_build) in `tox`, for a better developer experience

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,5 @@
+[build-system]
+requires = [
+  "setuptools>=50.3.2",
+]
+build-backend = "setuptools.build_meta"
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,7 @@
 
 [tox]
 envlist = py{36,37,38,39,py3}, linting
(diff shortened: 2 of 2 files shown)


ssbarnea (approved):

BeyondEvil:
What does this do and in what way does it lead to a better user experience @gnikonorov? 

Just trying to learn and understand. :) Read the tox docs, but that didn't give me a lot.

gnikonorov:
@BeyondEvil, it creates a separate isolated environment for your tox run. That way it's guaranteed that your local environment doesn't interfere in any way with the tox run.

BeyondEvil:
> @BeyondEvil, it creates a separate isolated environment for your tox run. That way it's guaranteed that your local environment doesn't interfere in any way with the tox run.

Ah, got it! Thanks! 👊

ssbarnea:
There is a cost involved as this makes tox slower but yep, improved isolation also good, that is why I approved it.

PR #351, opened 2020-10-18
Add Github Actions workflow for nodejs portion of project

gnikonorov:
Add in a github actions workflow for the `grunt` tests as they were missing.
 
Note that I used the following as a guide: https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-nodejs

The dependency list and command to run were taken directly out of the `.travis.yml` file: https://github.com/pytest-dev/pytest-html/blob/master/.travis.yml#L10. In a follow up PR I want to make a `package.json` file to clean this up a bit, but I am intentionally deferring it to introduce as little change as possible.


ssbarnea:
For some reason i do not see the new workflow triggered. It may be easier to manage if you add a new entry to the existing workflow file, so we do not repeat the conditions. Also this will enable us to put order conditions and likely solve the lack of triggering.

gnikonorov:
@ssbarnea - is there a way to test this before merge? I couldn't find anything except third party solutions to test locally. I'm confident in this change, but it would be nice to verify before reviewing/merging.

gnikonorov:
> For some reason i do not see the new workflow triggered. It may be easier to manage if you add a new entry to the existing workflow file, so we do not repeat the conditions. Also this will enable us to put order conditions and likely solve the lack of triggering.

I have to step out for a while, I'll try that when I'm available again

ssbarnea:
Modify existing workflow and it will work. Adding new workflow does not work.

gnikonorov:
Looks like it works as expected with your suggestion @ssbarnea. My only question is, do we want to rename the file in this PR? It is currently called `tox.yml` which is no longer accurate.

gnikonorov:
I renamed the file to `actions.yml`

BeyondEvil (approved):
Nice! Good job @gnikonorov !

Really appreciate the energy you've brought to the project this past week! 👍

gnikonorov:
Bump for a review @ssbarnea

ssbarnea (approved):

Issue #171, opened 2018-06-27
Don't include any captured stdout/stderr/log output to the report


bigunyak:
Hi,
To reduce report file size, I'm trying to exclude any captured stdout/stderr/log output.
I'm running tests with the following command:
```
pytest --html=report.html --self-contained-html --capture=no --show-capture=no tests/
```
With this I still get the `Captured log call` section for every test no matter if it pass or not.
So the question, how to not include any captured output to the report?

bigunyak:
The `--show-capture` option seems to have absolutely no effect on what get's included into the report.
Could you please help to understand how capturing suppose to work?

davehunt:
`--show-capture` is not related to the HTML report, but I like the idea. I'll label this as an enhancement and maybe somebody would be interested in working on it. I don't think it would be too hard, we'd just need to get the value from the configuration and make the writing to the report conditional based on it.

adriangb:
This honestly seems like a bug to me, not an enhancement.

BeyondEvil:
I agree with @adriangb.

What do you think @davehunt ?

rochacbruno:
what I understood is that `--capture` is a pytest argument not a `pytest-html` argument.

So the enhancement would be adding the feature to read that arguments and do something.

BeyondEvil:
I think it makes sense for the plugin to ”obey” the flag.

adriangb:
> what I understood is that `--capture` is a pytest argument not a `pytest-html` argument.
> 
> So the enhancement would be adding the feature to read that arguments and do something.

Yes, `--capture` and `--show-capture` are Pytest options, but regardless it would be nice to have these options in the plugin (there is a reason they exist in Pytest), so why not make the plugin follow whatever Pytest does? Then to go the extra mile `--html-report-capture` and `--html-report-show-capture` could be added which override `--capture` and `--show-capture` for the HTML report.

davehunt:
I still think this is an enhancement, and a worthwhile one. Perhaps pytest should be more clear that the capture arguments only relate to the console logging, and that plugins are not required to respect them. If you disagree then I would suggest opening a bug against pytest that prevents plugins from having access to this content if the arguments are used.

Either way, I don't think we want to duplicate these arguments for each plugin.

sjkcarnahan:
I would like to tack on that this would be a great option.

hello-ming:
Any progress here? It's better if we could control disabling stdout or stderr or all.

[...]

PR #353, opened 2020-10-21
Respect pytest --capture=no and -s flags

gnikonorov:
Make `pytest-html` respect the `--capture=no` and `-s` pytest flags as requested in https://github.com/pytest-dev/pytest-html/issues/171.

Note that this only partially resolves the issue. We should also respect `--show-capture=no`, but I will add that in a follow up pr to make the changes smaller.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -173,7 +173,7 @@ class HTMLReport:
             for extra_index, extra in enumerate(getattr(report, "extra", [])):
                 self.append_extra_html(extra, extra_index, test_index)
 
-            self.append_log_html(report, self.additional_html)
+            self.append_log_html(report, self.additional_html, config.option.capture)
 
             cells = [
                 html.td(self.outcome, class_="col-result"),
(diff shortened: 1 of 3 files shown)


pytest_html/plugin.py
```diff
@@ -278,41 +278,43 @@ def append_extra_html(self, extra, extra_index, test_index):
...
-                for line in text.splitlines():
-                    separator = line.startswith("_ " * 10)
-                    if separator:
-                        log.append(line[:80])
-                    else:
-                        exception = line.startswith("E   ")
-                        if exception:
-                            log.append(html.span(raw(escape(line)), class_="error"))
+
+            if pytest_capture_value != "no":
```
BeyondEvil, line 284: Since this `if` has no else or its own "body", you can probably do:

```python
if pytest_capture_value != "no" and report.longrepr:
```

BeyondEvil (approved):
This is beautiful @gnikonorov !

One small nitpick. 😊

pytest_html/plugin.py
gnikonorov: I don't think we could do that, since this `if` also contains the `for section in report.sections:` block. We'd have to duplicate the statement then which is confusing IMO

BeyondEvil:
Yes, you’re absolutely correct! I missed that the for-loop belonged to the outer if.

gnikonorov:
Bump for a review/merge @ssbarnea

gnikonorov:
Hey @BeyondEvil if you are OK with this PR can you merge it in ( I'd rather not merge my own PRs, since I don't know how everyone else feels about them ). I'd like this in master so I can start work on the second half of https://github.com/pytest-dev/pytest-html/issues/171, implementing support for `--show-capture`

ssbarnea (approved):

PR #354, opened 2020-10-21
Use os.makedirs(exist_ok=True).

glogiotatidis:
Replace os.path.exists checks with os.makedirs(exist_ok=True) which helps with race conditions when running tests with pytest-parallel or pytest-xdist.

### DIFF
--- a/pytest_html/plugin.py
+++ b/pytest_html/plugin.py
@@ -223,8 +223,7 @@ class HTMLReport:
                 os.path.dirname(self.logfile), "assets", asset_file_name
             )
 
-            if not os.path.exists(os.path.dirname(asset_path)):
-                os.makedirs(os.path.dirname(asset_path))
+            os.makedirs(os.path.dirname(asset_path), exist_ok=True)
 
             relative_path = f"assets/{asset_file_name}"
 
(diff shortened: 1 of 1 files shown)


gnikonorov (commented):
Thank you for the PR, @glogiotatidis!

I just want to confirm that this helps with race conditions due to the fact that we've consolidated the directory creation from 2 statements ( first check if the path doesn't exist, then make it ) into 1 ( make the path with `exists_ok` set to true )

gnikonorov:
Also pinging for a review from @ssbarnea or @BeyondEvil if either of you have time

glogiotatidis:
> I just want to confirm that this helps with race conditions due to the fact that we've consolidated the directory creation from 2 statements ( first check if the path doesn't exist, then make it ) into 1 ( make the path with `exists_ok` set to true )

Yes, that's the idea. Of course being a race condition it's pretty hard to replicate / validate but I believe this code would fix ti.

BeyondEvil:
This looks good @glogiotatidis !

I'd be great if you could add test(s) for this. I believe a test already exists where the directory does not exist, but not one where the directory already exists.

Thanks! 🙏

glogiotatidis:
Thanks @BeyondEvil. I'd actually go the other way and remove the existing test where the directory does not exist. The reason behind this is that now we're using build-in functionality to handle with (non)existing dirs and we don't really need to test this, as it's covered in python's test suite (or I hope it does :D)

BeyondEvil:
> Thanks @BeyondEvil. I'd actually go the other way and remove the existing test where the directory does not exist. The reason behind this is that now we're using build-in functionality to handle with (non)existing dirs and we don't really need to test this, as it's covered in python's test suite (or I hope it does :D)

Good point! 👍

BeyondEvil (approved):
LGTM 🚀

gnikonorov:
Thank you again @glogiotatidis! Merging this in

gnikonorov (approved):

alexgibson:
@gnikonorov Hi - any chance we could get this included in the 3.0.0 release? I'm not sure if it's already in there, but I didn't see it in https://github.com/pytest-dev/pytest-html/releases/tag/v3.0.0a0

ssbarnea:
It will be, we just need to sort the release process but no need to worry.

gnikonorov:
@alexgibson it will be included as @ssbarnea mentions above!

PR #357, opened 2020-10-23
Enable testing on Windows and MacOS

csm10495:
Hopefully the last try for this. @gnikonorov sorry about the roundabouts.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -76,9 +76,10 @@ jobs:
             name: pypy3-ubuntu
             python-version: pypy3
 
-          - os: windows-latest
-            name: pypy3-windows
-            python-version: pypy3
+          # Disabling due to https://github.com/actions/setup-python/issues/151
+          # - os: windows-latest
+          #   name: pypy3-windows
+          #   python-version: pypy3
 
           - os: macOS-latest
             name: pypy3-mac
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -16,25 +16,82 @@ on:
 jobs:
   build_python:
     name: ${{ matrix.name }}
-    runs-on: ubuntu-18.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         include:
-          - name: linting
+          - os: ubuntu-18.04
+            name: linting-ubuntu
             python-version: 3.6
-          - name: py36
+
+          - os: ubuntu-18.04
+            name: py36-ubuntu
             python-version: 3.6
-          - name: py37
+
+          - os: windows-latest
+            name: py36-windows
+            python-version: 3.6
+
+          - os: macOS-latest
+            name: py36-mac
+            python-version: 3.6
+
+          - os: ubuntu-18.04
+            name: py37-ubuntu
             python-version: 3.7
-          - name: py38
+
+          - os: windows-latest
+            name: py37-windows
+            python-version: 3.7
+
+          - os: macOS-latest
+            name: py37-mac
+            python-version: 3.7
+
+          - os: ubuntu-18.04
+            name: py38-ubuntu
             python-version: 3.8
-          - name: py39
+
+          - os: windows-latest
+            name: py38-windows
+            python-version: 3.8
+
+          - os: macOS-latest
+            name: py38-mac
+            python-version: 3.8
+
+          - os: ubuntu-18.04
+            name: py39-ubuntu
             python-version: 3.9
-          - name: devel
-            python-version: 3.8
-          - name: pypy3
+
+          - os: windows-latest
+            name: py39-windows
+            python-version: 3.9
+
+          - os: macOS-latest
+            name: py39-mac
+            python-version: 3.9
+
+          - os: ubuntu-18.04
+            name: pypy3-ubuntu
             python-version: pypy3
+
+          - os: windows-latest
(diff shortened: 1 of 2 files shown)


csm10495:
Looks all good. Someone may just need to change the jobs gh actions is looking for since i changed the names.

gnikonorov:
> Hopefully the last try for this. @gnikonorov sorry about the roundabouts.

Thank you! No worries at all @csm10495

testing/test_pytest_html.py
```diff
@@ -12,6 +13,21 @@
 
 pytest_plugins = ("pytester",)
 
+# Force a utf-8 encoding on file io. See
```
gnikonorov, line 16: since this is only for windows can we conditionally execute this if we detect we are on windows ( with [sys.platform](https://docs.python.org/3/library/sys.html#sys.platform) for example ). I'd prefer to avoid modifying builtins unless we have to

.github/workflows/actions.yml
```diff
@@ -16,25 +16,83 @@ on:
...
+
+          - os: macOS-latest
+            name: py39-mac
+            python-version: 3.9
+
+          - os: ubuntu-18.04
+            name: pypy3-ubuntu
             python-version: pypy3
+
+          # Disabling due to https://github.com/actions/setup-python/issues/151
```
gnikonorov, line 79: Can you raise an issue to reintroduce this when https://github.com/actions/setup-python/issues/151 is resolved?

.github/workflows/actions.yml
```diff
@@ -43,9 +101,15 @@ jobs:
     - name: Install tox
       run: |
         python -m pip install --upgrade tox
+    - name: Get Tox Environment Name From Matrix Name
+      uses: rishabhgupta/split-by@v1
+      id: split
```
gnikonorov, line 106: nitpick: can we give this a more verbose name? Something like `split-matrix-name`

gnikonorov (changes requested):
Thank you for this!

I have a few small comments. Also I noticed that the github actions page has the following for windows:
```
WARNING: The scripts pip.exe, pip3.9.exe and pip3.exe are installed in 'C:\hostedtoolcache\windows\Python\3.9.0\x64\Scripts' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
```

Could we resolve this as well? This is the URL: https://github.com/pytest-dev/pytest-html/actions/runs/323299476. If you scroll down to  the 'Annotations' section you should see it

csm10495:
> Thank you for this!
> 
> I have a few small comments. Also I noticed that the github actions page has the following for windows:
> 
> ```
> WARNING: The scripts pip.exe, pip3.9.exe and pip3.exe are installed in 'C:\hostedtoolcache\windows\Python\3.9.0\x64\Scripts' which is not on PATH.
>   Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
> ```
> 
> Could we resolve this as well? This is the URL: https://github.com/pytest-dev/pytest-html/actions/runs/323299476. If you scroll down to the 'Annotations' section you should see it

I'm not sure how to resolve that one. They have an open issue for it here: https://github.com/actions/setup-python/issues/146.

The other stuff I'll look into now

.github/workflows/actions.yml
csm10495: yep: https://github.com/pytest-dev/pytest-html/issues/358

.github/workflows/actions.yml
csm10495: sure

testing/test_pytest_html.py
csm10495: Yup!

csm10495:
Updated to fix the small things.

gnikonorov:
Thank you @csm10495! It's late here, I'll take a look again tomorrow.

gnikonorov (approved):
Thanks @csm10495! This LGTM.

I'll defer merging and final approval to @ssbarnea, since he is more knowledgeable in github workflows than me

.github/workflows/actions.yml
```diff
@@ -43,9 +101,15 @@ jobs:
     - name: Install tox
       run: |
         python -m pip install --upgrade tox
+    - name: Get Tox Environment Name From Matrix Name
```
ssbarnea, line 104: I do not fancy depending on other repositories for a single liner. I know is should be safe due to due use of tags but still prefer to have less dependencies.

I found more practical to do a little bit of duplication and add an explicit `tox_env` variable defined in each matrix element. It adds few lines but makes is much easier to maintain.

testing/test_pytest_html.py
```diff
@@ -12,6 +13,20 @@
 
 pytest_plugins = ("pytester",)
 
+if os.name == "nt":
```
ssbarnea, line 16: If I remember way that is not the ideal way to test for windows as it produces different results based on which kind of python you run with. My impression is that platform.system() is more reliable: https://github.com/pycontribs/subprocess-tee/blob/main/lib/subprocess_tee/__init__.py#L22

ssbarnea (approved):

ssbarnea:
There are few things I do not like about this change but I will merge it because I do not want demoralize @csm10495 --- we can look at addressing my comments in follow-ups.

PR #360, opened 2020-10-24
Move release from travis to github actions

ssbarnea:
- create packaging environment that produces deliverables
- removes travis integration
- configures github actions to call packaging after all the other
  tests are passing and to make a release if tag is present
- test.pypi.org is kept commented until we sort credentials for it
- no credentials are needed as they are already configured in github
- enables a daily cron run of CI, so we know if something goes wrong for external reasons


ssbarnea:
@BeyondEvil @gnikonorov That is ready to ship!

BeyondEvil:
I've started the review. But since the GHA is all new to me, it's going to take a while.

.github/workflows/actions.yml
```diff
@@ -12,6 +12,8 @@ on:
     - >-
       **
   pull_request:
+  schedule:
```
gnikonorov, line 15: You're just introducing a daily CI build here, right?

.github/workflows/actions.yml
```diff
@@ -127,3 +129,71 @@ jobs:
...
+          # history for them.
+          fetch-depth: >-
+            ${{
+              (
+                github.event_name == 'create' &&
+                github.event.ref_type == 'tag'
+              ) &&
+              1 || 0
+            }}
+      - name: Drop Git tags from HEAD for non-tag-create events
```
gnikonorov, line 166: just out of curiosity, why do we want to drop all tags for non tag create events?

gnikonorov (commented):
I did a quick skim now. Overall I think it looks good, I just had one question out of curiosity. I have to step away, but I'll do a proper review when I'm back.

.github/workflows/actions.yml
```diff
@@ -127,3 +129,71 @@ jobs:
...
+        run: python -m pip install --user tox
+      - name: Check out src from Git
+        uses: actions/checkout@v2
+        with:
+          # Get shallow Git history (default) for tag creation events
+          # but have a complete clone for any other workflows.
+          # Both options fetch tags but since we're going to remove
+          # one from HEAD in non-create-tag workflows, we need full
+          # history for them.
+          fetch-depth: >-
```
nicoddemus, line 158: Also not clear to me why we are deleting tags here, but I think just setting this to `0` is OK: pytest-html is a tiny repo so cloning the entire history is fine IMHO.

nicoddemus (commented):
Agree, overall looks good, I have the same questions regarding deleting tags though.

gnikonorov (changes requested):
Since this will officially deprecate our usage of travis, can you remove any references to it as part of this commit or open an issue after this merges so we don't forget? I see the following references to it in the codebase:

```
$ git grep travis
README.rst:.. image:: https://img.shields.io/travis/pytest-dev/pytest-html.svg
README.rst:   :target: https://travis-ci.org/pytest-dev/pytest-html/
development.rst:All pull requests and merges are tested in `Travis CI <https://travis-ci.org/>`_
development.rst:based on the ``.travis.yml`` file.
development.rst:Usually, a link to your specific travis build appears in pull requests, but if
development.rst:`pull requests page <https://travis-ci.org/pytest-dev/pytest-html/pull_requests>`_
development.rst:8. Done. You can monitor the progress on `Travis <https://travis-ci.org/pytest-dev/pytest-html/>`_
$
```

.github/workflows/actions.yml
ssbarnea: Exactly.

.github/workflows/actions.yml
ssbarnea: I do not remember exactly as I was not the original author of this task, maybe @webknjaz could help here?

My guess is to avoid getting setuptools-scm confused about which tag it would use for producing the version, especially as it is possible to add a tag somewhere back-in-time (before more recent tag in the branch).

ssbarnea:
> Since this will officially deprecate our usage of travis, can you remove any references to it as part of this commit or open an issue after this merges so we don't forget? I see the following references to it in the codebase:
> 
> ```
> $ git grep travis
> README.rst:.. image:: https://img.shields.io/travis/pytest-dev/pytest-html.svg
> README.rst:   :target: https://travis-ci.org/pytest-dev/pytest-html/
> development.rst:All pull requests and merges are tested in `Travis CI <https://travis-ci.org/>`_
> development.rst:based on the ``.travis.yml`` file.
> development.rst:Usually, a link to your specific travis build appears in pull requests, but if
> development.rst:`pull requests page <https://travis-ci.org/pytest-dev/pytest-html/pull_requests>`_
> development.rst:8. Done. You can monitor the progress on `Travis <https://travis-ci.org/pytest-dev/pytest-html/>`_
> $
> ```

Indeed, I forgot the cleanup travis references. I will update the PR.

.github/workflows/actions.yml
```diff
@@ -127,3 +129,71 @@ jobs:
...
+          git tag --points-at HEAD
+          |
+          xargs git tag --delete
+      - name: Build dists
+        run: python -m tox
+      # Disabled until we get pytest-dev account maintainer on test.pypi.org
+      # - name: Publish to test.pypi.org
+      #   if: >-
+      #     (
+      #       github.event_name == 'push' &&
```
webknjaz, line 180: How about
```suggestion
      #       false &&
      #       github.event_name == 'push' &&
```
instead of commenting out the whole thing?

.github/workflows/actions.yml
webknjaz: @ssbarnea yes, that would be one subcase.
Although, I've been doing this in particular to avoid a slightly different race condition.
When the tags are pushed, this generates 2 workflows: for tag `push` and for tag `create`.
This means that whenever both of these workflows will run at about the same time and the checkout action will get the tag and the commits. And then, setuptools-scm will generate a version with that tag exactly. And this is fine because later in the workflow, the publish step is executed under only one of these cases so both jobs won't attempt to upload. That's why this step is skipped for these cases.
Now, there's a third job that may run separately in the context of the same commit: `push` to master. And if that commit is also tagged (for example, if the tagging happens fast after pushing the same commit to master, before some of the jobs did checkout), setuptools-scm will generate a version based on that tag. This is bad when that "Publish to test.pypi.org" step is uncommented because it'll result in two jobs attempting to publish an artifact with the same name to TestPyPI and one of them will fail (with no guarantee which one). And so for this case, I'm removing any tags that may be present in the HEAD which eliminates the race condition.

tox.ini
```diff
@@ -34,6 +34,34 @@ deps =
...
+[testenv:packaging]
+description =
+    Do packaging/distribution. If tag is not present or PEP440 compliant upload to
+    PYPI could fail
+# `usedevelop = true` overrides `skip_install` instruction, it's unwanted
+usedevelop = false
+# don't install package in this env
+skip_install = true
+deps =
+    collective.checkdocs >= 0.2
```
webknjaz, line 46: Isn't this used for `setup.py check` that is not executed here?

tox.ini
```diff
@@ -34,6 +34,34 @@ deps =
...
+    pep517 >= 0.8.2
+    pip >= 20.2.2
+    toml >= 0.10.1
```
webknjaz, line 49: This is a transitive dependency of `pep517`, it probably doesn't matter what version it is.

webknjaz:
FTR these days I'd rely on https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/ instead of `create` because `create` requires post-checks in runtime and also runs on branch creation too.

.github/workflows/actions.yml
gnikonorov: Thank you for clarifying!

.github/workflows/actions.yml
gnikonorov: Thank you for clarifying @webknjaz. Since @nicoddemus also had the same question, I'll ask him to resolve this if this answer works for him

development.rst
```diff
@@ -23,15 +23,12 @@ If you're not using `Pipenv`_, to install `pre-commit`, run:
 Automated Testing
 -----------------
 
-All pull requests and merges are tested in `Travis CI <https://travis-ci.org/>`_
-based on the ``.travis.yml`` file.
+All pull requests and merges are tested in `GitHub Actions <https://github.com/pytest-dev/pytest-html/actions>`_
+which are defined inside ``.github`` folder.
```
gnikonorov, line 27: nitpick: 'inside the ``.github`` folder.'

development.rst
```diff
@@ -23,15 +23,12 @@ If you're not using `Pipenv`_, to install `pre-commit`, run:
...
 
-Usually, a link to your specific travis build appears in pull requests, but if
-not, you can find it on the
-`pull requests page <https://travis-ci.org/pytest-dev/pytest-html/pull_requests>`_
-
-The only way to trigger Travis CI to run again for a pull request, is to submit
-another change to the pull branch.
+To retrigger CI to run again for a pull request, you can use either use
+dropdown option, close and reopen pull-request or to just update the branch
```
gnikonorov, line 30: nitpick: 'use the dropdown option, close and reopen the pull-request or just update the branch'

gnikonorov (changes requested):
Thank you for cleaning up the docs! I have a few nitpicks, but don't mind approving after @webknjaz's feedback is addressed.

webknjaz:
FWIW none of my feedback is blocking, it's rather informative.

gnikonorov:
> FWIW none of my feedback is blocking, it's rather informative.

Thank you for clarifying @webknjaz! 

@ssbarnea let me know then when you've addressed all the feedback you wanted to and I'll approve. I'd like to wait for @BeyondEvil to also approve before we merge

.github/workflows/actions.yml
nicoddemus: Thanks @webknjaz for the explanation.

Under normal circumstances in our release workflow the race condition cannot really happen, because the tag will be created in the last commit of the release PR, and only after that is green (and hence the package is published), the commit will be merged to `master`, not as fast-forward but as a normal merge commit.

So I'm under the opinion that this is a bit of complication that really doesn't bring anything to the table for `pytest-dev/pytest-html`, but it is also harmless so feel free to go ahead with it. 👍

nicoddemus:
> FTR these days I'd rely on github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch instead of create because create requires post-checks in runtime and also runs on branch creation too.

I also think this would be nice, but just as @webknjaz, this not blocking for me. 😁

nicoddemus (approved):

ssbarnea:
I managed to setup testpypi and I will update this PR tomorrow.

gnikonorov:
Hey @ssbarnea can we merge this in first, and then add `testpypi` in a follow up PR?

tox.ini
ssbarnea: Added

ssbarnea:
I had to rebase, fixed additional commend. To avoid delaying this even more lets merge it as soon we get the review and we will do other improvements in follow-ups.

gnikonorov:
> I had to rebase, fixed additional commend. To avoid delaying this even more lets merge it as soon we get the review and we will do other improvements in follow-ups.

Of course @ssbarnea! I'll take a look today. Thank you

gnikonorov (approved):
LGTM!

As @ssbarnea mentioned we can get any improvements we need in subsequent PRs. I think this is fine as is and will let us start releasing with the new workflow.

I would like one more person to approve before merge, however. @BeyondEvil  can you take a look as well?

.github/workflows/actions.yml
```diff
@@ -127,3 +129,70 @@ jobs:
...
+
+    env:
+      PY_COLORS: 1
+      TOXENV: packaging
+
+    steps:
+      - name: Switch to using Python 3.6 by default
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
```
BeyondEvil, line 147: Is there a reason why we're using 3.6 and not a newer version @ssbarnea @gnikonorov ?

webknjaz: I'm pretty sure Sorin wanted to support the oldest possible version that the devs/contributors may use.

ssbarnea: Yes, use of oldest supported is usually better than newest in order to avoid regressions. Based on experience using oldest proved to catch more issues than the newest, especially in this area. 

Any CI failure that prevents a bad release is better than making a user unhappy. Note to myself: add comments on action tasks like this documenting their purpose.

BeyondEvil: Thanks @ssbarnea! Makes sense. 👍

Issue #355, opened 2020-10-22
Clarifying the release schedule


gnikonorov:
Hi everyone.

I think it may make sense to create a release schedule. I looked at PyPI and it looks like we haven't released a new version of this plugin since March of this year. 

While we're at it, perhaps it would make sense to automate the process as much as possible and create a `RELEASING.rst` doc similar to [pytest's](https://github.com/pytest-dev/pytest/blob/master/RELEASING.rst). This way users could be aware of what to expect in terms of feature delivery.

Thoughts @ssbarnea @BeyondEvil @davehunt ?

BeyondEvil:
I'm no fan of release schedules for relatively small projects. They make sense for larger projects that have a roadmap.

As for "automatic" releases, I highly recommend [Semantic Release](https://github.com/semantic-release/semantic-release) we use it heavily at work (and it's not only for npm projects).

gnikonorov:
> I'm no fan of release schedules for relatively small projects. They make sense for larger projects that have a roadmap.

You're right. It probably doesn't make sense for this plugin now that I think about it.

> As for "automatic" releases, I highly recommend [Semantic Release](https://github.com/semantic-release/semantic-release) we use it heavily at work (and it's not only for npm projects).

This looks awesome! So it would be part of our pipeline, and then a separate step could publish the build product to PyPI right?

Also, can we publish a package now in the meantime since it's been a while and the changelog has a few unreleased items in it? I would do it, but I don't think I have the permission to.

BeyondEvil:
> > As for "automatic" releases, I highly recommend [Semantic Release](https://github.com/semantic-release/semantic-release) we use it heavily at work (and it's not only for npm projects).
> 
> This looks awesome! So it would be part of our pipeline, and then a separate step could publish the build product to PyPI right?

So, high level, this is how it works:

On every merge to the default branch (in our case `master`) Semantic Release is run. If it determines a release can be made (by scanning all the commits since last tag) it will create and push a new tag.

This tag will then trigger a workflow which publishes a package to PyPi.

Since I have experience with this, I'm happy to set it up if it makes sense to you and @ssbarnea.

> Also, can we publish a package now in the meantime since it's been a while and the changelog has a few unreleased items in it? I would do it, but I don't think I have the permission to.

Sure! You can follow the steps [here](https://github.com/pytest-dev/pytest-html/blob/master/development.rst#releasing-a-new-version) and I help you if you get stuck. 👍

[...]

PR #362, opened 2020-10-25
Prep version 3.0.0 for release

gnikonorov:
Prep version 3.0.0 for release as discussed in https://github.com/pytest-dev/pytest-html/issues/355. Major bump is due to https://github.com/pytest-dev/pytest-html/pull/347

I plan on releasing a new version the same day this PR merges

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Release Notes
 -------------
 
-**2.1.2 (unreleased)**
+**3.0.0 (2020-10-25)**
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

ssbarnea:
Merging with the remark that: do not create the release until we merge the workflow changes, so we can validate them too. I will create first create a `3.0.0a0` pre-release/tag, and if that works fine we convert it to full release.

Issue #322, opened 2020-07-30
Replace travis with github-actions


ssbarnea:
I created this in because I want to exemplify why I think in 2020, it makes more sense to use Github Action than Travis.

There is nothing preventing us from running both in parallel for a longer period of time, and even ignoring the results from one of them if we want.

I can easily raise a PTs I already did this for lots of repos and I know the tricks.

## Update 2020-10

Changed scope from trying... to switching...

BeyondEvil:
> I created this in because I want to exemplify why I think in 2020, it makes more sense to use Github Action than Travis.
> 
> There is nothing preventing us from running both in parallel for a longer period of time, and even ignoring the results from one of them if we want.
> 
> I can easily raise a PTs I already did this for lots of repos and I know the tricks.

I would really appreciate a PR with some Github Actions! 😊 

It would be awesome if you could comment it extensively to expedite the learning curve for me (and future contributors). Especially any tricks. 👍

gnikonorov:
+1 to @BeyondEvil's response

ssbarnea:
I think we all seen how nice github actions works. Maybe is time to effectively drop Travis and switch the release to github actions. Less complexity can only help the project.

@BeyondEvil @davehunt What do you think?

I cannot do it myself because I do not have access to project settings or pypi. One needs to add project-specific-tokens for both pypi instances that look like:

![](https://sbarnea.com/ss/Screen-Shot-2020-10-17-13-35-46.98.png)

Ideally we should use a bot account for that, but I see none configured. You could reuse the `pytest-dev` bot for that.

gnikonorov:
Regarding the above, I think we should drop travis and move to github actions

BeyondEvil:
I'm happy to make the switch over to only using GHA. I also think using the pytest-dev is a good idea as long as project owners have the access they need to it.

With that said, @davehunt has the final say.

davehunt:
> With that said, @davehunt has the final say.

Don't let me get in the way of the using the best available tools. If you're all happy, I'm certainly not going to stop you.

BeyondEvil:
Where can I see the pytest-dev in use with GHA? @ssbarnea

BeyondEvil:
I need your pypi username @ssbarnea

What's the username of the pytest-dev bot @nicoddemus ? Will I have access to it? Do I need access to it (to make it the release bot)?

[...]

PR #364, opened 2020-11-04
Assure scm versioning is pypa compatible

ssbarnea:
This fixes the issue of releasing test.pypi.org containers that were not tagged by making the versioning compatible with publishing requirements (taken from other places where I am using it, mainly avoids using ``+`` in version and instead produces something like ``2.1.2.dev47`` which is accepted.

Related: https://github.com/pytest-dev/pytest-html/issues/322#issuecomment-721315174

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,12 @@
 [build-system]
 requires = [
-  "setuptools>=50.3.2",
+  "pip >= 19.3.1",
+  "setuptools >= 42",
+  "setuptools_scm[toml] >= 3.5.0",
+  "setuptools_scm_git_archive >= 1.1",
+  "wheel >= 0.33.6",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
+local_scheme = "no-local-version"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,5 +1,12 @@
 [build-system]
 requires = [
-  "setuptools>=50.3.2",
+  "pip >= 19.3.1",
+  "setuptools >= 42",
+  "setuptools_scm[toml] >= 3.5.0",
+  "setuptools_scm_git_archive >= 1.1",
```
gnikonorov, line 6: Out of curiosity, why do we need to add all these requirements for this change?

gnikonorov (approved):
Looks good to me, but can you please answer my question about the additions to `requires` in `pyproject.toml`?

ssbarnea:
TBH, I am not sure if all these are required but if they are not yet, they will soon be. Based on how pip is called and its own version it can decide to use very different methods for installing. This addresses some other issues people may encounter. Keep in mind that these are build time requirements, something very different that runtime requirements.

PR #365, opened 2020-11-07
Add a package.json file for a better javascript development experience

gnikonorov:
Add a `package.json` file to simplify the experience of developing the JS portion of this plugin.

This also better sets us up to tackle https://github.com/pytest-dev/pytest-html/issues/350 since we won't have to remember all the JS dependencies anymore, and can use more modern practices for JS development.


package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "pytest-html",
+  "version": "3.0.0",
```
ssbarnea, line 3: My experience with nodejs is low but I would really try to avoid having an extra new place to update the versions. The code should not keep the version, this should be takes from git (scm) tags, like setuptools-scm does.

gnikonorov: @ssbarnea I can try removing it and seeing if the install and test target still run fine. I don't see a need to have one since we're not publishing to npm. 

I'll try when I'm free this afternoon.

gnikonorov:
@ssbarnea I ended up removing all data from the `package.json` that was related to publishing, so now it only has what is necessary to set up the JavaScript environment for and run the tests. I think it is good for another look when you have a chance.

gnikonorov:
Bump @ssbarnea or @BeyondEvil

package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "main": "Gruntfile.js",
+  "dependencies": {
+    "grunt": "^1.3.0",
+    "grunt-cli": "^1.3.2",
+    "grunt-contrib-qunit": "^4.0.0",
+    "phantomjs-prebuilt": "2.1.15"
+  },
+  "devDependencies": {},
```
gnikonorov, line 9: Note that I intentionally left this in. I'd rather be explicit in listing that there are no dev dependencies than have people guessing.

development.rst
```diff
@@ -52,18 +52,38 @@ Otherwise, to install and run, do:
...
+1.  Update your local master with the upstream master (``git pull --rebase upstream master``)
+2.  Create a new branch
+3.  Update ``CHANGES.rst`` with the new version, today's date, and all changes/new features
+4.  Update the ``version`` field in  ``package.json`` with the new version
+5.  Commit and push the new branch and then create a new pull request
+6.  Wait for tests and reviews and then merge the branch
+7.  Once merged, update your local master again (``git pull --rebase upstream master``)
+8.  Tag the release with the new release version (``git tag v<new tag>``)
+9.  Push the tag (``git push upstream --tags``)
+10. Done. Check `CI <https://github.com/pytest-dev/pytest-html/actions>`_ for release progress.
```
BeyondEvil, line 87: Here's a trick: You can change all the numbers to `1`, and RST will renumber them. That way, if you have to insert something, it will only be a one-line change. :)

BeyondEvil (approved):
LGTM! One small comment.

ssbarnea (approved):

PR #367, opened 2020-11-11
Add ESLint to project

gnikonorov:
Add [ESLint](https://eslint.org/) for JS linting.

This better sets us up to tackle https://github.com/pytest-dev/pytest-html/issues/350, and should be added in as a best practice regardless.

The `.eslintrc.json` file was auto-generated by `ESLint`.


gnikonorov:
Note the `grunt` task is not actually run anymore. It's only showing because I didn't rename the action in my initial PR. It's an issue with GitHub actions and not the change

ssbarnea (changes requested):
You missed the most important bit: adding eslint to .pre-commit-config.yaml so it runs during linting.

If you do this you do not need to alter any github actions.

pytest_html/resources/main.js
```diff
@@ -182,15 +181,15 @@ function key_num(col_index) {
 
 function key_link(col_index) {
     return function(elem) {
-        dataCell = elem.childNodes[1].childNodes[col_index].firstChild
-        return dataCell == null ? "" : dataCell.innerText.toLowerCase();
+        var dataCell = elem.childNodes[1].childNodes[col_index].firstChild;
```
BeyondEvil, line 184: this should probably be `const`

pytest_html/resources/main.js
```diff
@@ -200,12 +199,12 @@ function reset_sort_headers() {
         elem.parentNode.removeChild(elem);
     });
     find_all('.sortable').forEach(function(elem) {
-        var icon = document.createElement("div");
-        icon.className = "sort-icon";
-        icon.textContent = "vvv";
+        var icon = document.createElement('div');
```
BeyondEvil, line 202: `const`

BeyondEvil (changes requested):
Almost every `var` can be changed to `const`. 

Add this rule: https://eslint.org/docs/rules/prefer-const

And it should help you out.

gnikonorov:
> Almost every `var` can be changed to `const`. 
> 
> 
> 
> Add this rule: https://eslint.org/docs/rules/prefer-const
> 
> 
> 
> And it should help you out.

Let me take some time today and read over the rule list to see what else we may want to add

gnikonorov:
> You missed the most important bit: adding eslint to .pre-commit-config.yaml so it runs during linting.
> 
> If you do this you do not need to alter any github actions.

Good point @ssbarnea, thank you for pointing this out. I've updated the PR accordingly

gnikonorov:
@BeyondEvil @ssbarnea I've gone through the rule list and added every rule I thought made sense that didn't involve sweeping changes ( I'd feel safer adding those in a follow up PR to avoid breaking things ) except for the following sections:
- https://eslint.org/docs/rules/#stylistic-issues
- https://eslint.org/docs/rules/#ecmascript-6

From the rest of the rule list, I intentionally skipped the following rules since they involved making a refactor, which I want to address in a follow up PR to decrease the size of this one:
- https://eslint.org/docs/rules/no-use-before-define
- https://eslint.org/docs/rules/#strict-mode
- https://eslint.org/docs/rules/camelcase

If we are happy with this PR, I'd like to raise the above as separate issues after this merges so we don't forget about them and add them in time permitting.

BeyondEvil (approved):
LGTM! 🎉 

But let's wait for @ssbarnea before merging.

ssbarnea (approved):

PR #370, opened 2020-11-14
Add codecov support for python code

gnikonorov:
This PR adds coverage reporting and publishing to [codecov](https://codecov.io/)

This will better set us up to tackle https://github.com/pytest-dev/pytest-html/issues/350 as we'll have more confidence in the changes we're making.

Coverage is outputted to the screen and written to an xml file as the screen output assists developers and the xml file is actually uploaded to codecov.

FWIW, it turns out we have pretty solid python code coverage:
```
----------- coverage: platform linux, python 3.7.5-final-0 -----------
Name                                                            Stmts   Miss     Cover   Missing
------------------------------------------------------------------------------------------------
.tox/py37/lib/python3.7/site-packages/pytest_html/__init__.py       7      2    71.43%   7-9
.tox/py37/lib/python3.7/site-packages/pytest_html/extras.py        28      0   100.00%
.tox/py37/lib/python3.7/site-packages/pytest_html/hooks.py          5      0   100.00%
.tox/py37/lib/python3.7/site-packages/pytest_html/plugin.py       380      8    97.89%   36-38, 253, 295, 353, 418-419, 621
------------------------------------------------------------------------------------------------
TOTAL                                                             420     10    97.62%
```

I will add `JavaScript` as a follow up. Integrating codecov and [QUnit](https://qunitjs.com/) is not as straightforward and could even involve some refactoring. I'll create an issue for it once this merges.

Note that I moved the linting matrix entry of `build_python` into its own job for the following reasons:
- It was breaking `codecov` uploads since linting doesn't generate coverage files
- It wasn't accurate to have it under the python matrix since it also checks the `JavaScript` and `rst` files via the `eslint` and `rst` `precommit` actions respectively

Also note that `linting-ubuntu` will never report its status, since it's been replaced by `linting`

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[coverage:report]
+show_missing = true
+precision = 2
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,3 @@
 [coverage:report]
-show_missing = true
 precision = 2
+show_missing = true
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,63 @@
+{
+    "env": {
+        "browser": true,
+        "es2021": true
+    },
+    "extends": "eslint:recommended",
+    "parserOptions": {
+        "ecmaVersion": 12
+    },
+    "rules": {
+        "array-bracket-spacing": "error",
+        "block-scoped-var": "error",
+        "block-spacing": "error",
+        "brace-style": "error",
+        "class-methods-use-this": "error",
+        "consistent-return": "error",
+        "default-case": "error",
+        "default-case-last": "error",
+        "default-param-last": "error",
+        "grouped-accessor-pairs": "error",
+        "indent": [
+            "error",
+            4
+        ],
+        "linebreak-style": [
+            "error",
+            "unix"
+        ],
+        "no-caller": "error",
+        "no-console": "error",
+        "no-empty-function": "error",
+        "no-eval": "error",
+        "no-extra-parens": "error",
+        "no-labels": "error",
+        "no-new": "error",
+        "no-new-func": "error",
+        "no-new-wrappers": "error",
+        "no-return-await": "error",
+        "no-script-url": "error",
+        "no-self-compare": "error",
+        "no-shadow": "error",
+        "no-throw-literal": "error",
+        "no-undefined": "error",
+        "no-unreachable-loop": "error",
+        "no-unused-expressions": "error",
+        "no-useless-backreference": "error",
+        "no-useless-concat": "error",
+        "no-var": "error",
+        "prefer-const": "error",
+        "prefer-promise-reject-errors": "error",
+        "require-atomic-updates": "error",
+        "require-await": "error",
+        "semi": [
+            "error",
+            "always"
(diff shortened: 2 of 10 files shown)


gnikonorov:
For those that are curious here is this PRs coverage report for python: https://codecov.io/gh/pytest-dev/pytest-html/branch/add_codecov_support

I suspect https://codecov.io/gh/pytest-dev/pytest-html will populate once a PR with codecov support(  like this one ) merges into master

ssbarnea (changes requested):
I personally do not want to see a terminal report of coverage when I run the tests. This adds extra lines I need to scroll to see any test results.

There were serious discussions around removal of reporting from pytest-cov itself due to multiple design issues.

Coverage is good, especially when is done in a consolidated level using CI but it should never interfere with normal test executions.

So please keep coverage optional, and not enabled by a simple "tox -e py36", or even a simple "pytest" from the project root. The use may not even have pytest-cov installed and that is ok (hint avoid adding extra cli arguments).

There are multiple ways to achieve that: make use of PYTEST_OPTS and define it at ci level, or adding new set of tox environments that do also perform coverage like "py36-cov" kind of. No need to create new sections, you can add `-cov: ...` selector to alter what happens when running with coverage enabled.

Example:

deps =
  fooo
  -cov: pytest-cov

Something like this would install pytest-cov only in envs like py36-cov and not in simple py36 ones.

After you get tox working locally we can update CI to run the coverage variants.

gnikonorov:
@ssbarnea sure let me have coverage optional.

I haven't heard of purest-cov going away. At least it looks like the pytest-dev repos still use it. Is it still actively being considered for retirement?

ssbarnea:
It is about removal of `reporting` from it, not the coverage itself. That is the big issue: if you try to produce any reports while running the tests. Ideally it should only produce coverage files and zero reports. Reporting should be done separately, likely even a different tox env to collect all reports for local runs or codecov for ci. Still, all outside normal tox run.

gnikonorov:
How do you feel about making a separate env, changing CI to use this new env, and keeping the reporting in this pass. Then we can investigate reporting alternatives if we need to. It's better than what we have now.

@ssbarnea

ssbarnea (approved):
While not still perfect at least is not enforcing coverage when I run pytest from command line. We can look into improving it later, if the report becomes too annoying in the console.

gnikonorov:
Thank you @ssbarnea for confirming.

Let me go ahead and make the change

gnikonorov:
Failure is due to the fact that `devel-cov` is not recognized.

I have to step out but I'll take a look when I'm back

gnikonorov:
Sorry for the delay @BeyondEvil / @ssbarnea. This should be good to merge now if you two are OK with the change

gnikonorov:
Bump for a review @ssbarnea / @BeyondEvil

BeyondEvil (approved):
LGTM! 👍

gnikonorov:
Thanks @BeyondEvil ! @ssbarnea did you want to take another look or is this good to merge?

ssbarnea (approved):

PR #371, opened 2020-11-15
Remove outdated release instruction

gnikonorov:
This removes the step in the release notes on updating the `package.json` file, which is outdated. [The file](https://github.com/pytest-dev/pytest-html/blob/master/package.json) does not have a `version` field anymore, so this is confusing.

I'm also rewriting the release instruction list as an RST [ordered list](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/lists.html#ordered-lists), as it'll be easier to maintain than using numbers. If we use numbers and don't number the list correctly it won't format correctly whereas using the pound sign is always guaranteed to format properly.

### DIFF
--- a/development.rst
+++ b/development.rst
@@ -75,15 +75,14 @@ Releasing a new version
 
 Follow these steps to release a new version of the project:
 
-1.  Update your local master with the upstream master (``git pull --rebase upstream master``)
-2.  Create a new branch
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):
As long list is going down, I am for it!

PR #379, opened 2020-11-25
Add codecov badge to readme

gnikonorov:
Add a coverage badge to the `README.rst` file now that we have codecov up. The link was recommended by codecov in the Settings -> Badge  section under 'RST'

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,6 +22,9 @@ HTML report for the test results.
 .. image:: https://img.shields.io/requires/github/pytest-dev/pytest-html.svg
    :target: https://requires.io/github/pytest-dev/pytest-html/requirements/?branch=master
    :alt: Requirements
+.. image:: https://codecov.io/gh/pytest-dev/pytest-html/branch/master/graph/badge.svg?token=Y0myNKkdbi
+   :target: https://codecov.io/gh/pytest-dev/pytest-html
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
We don't need more badges, it will break rendering on some screens by wrapping the badge line. Example: https://pypi.org/project/pytest-html/

ssbarnea:
PS. I could change my mind if we remove another one. I have nothing against coverage badge in particular, only about total number of badges. I hate to see wrapped badge lines, on github or pypi. 

PS. I use split screens, do not assume everyone uses landscape mode.

gnikonorov:
Sure. How about we remove the 'open issues' badge? I think it gives less value than a codecov badge since it just shows the same information as the issues tab.

ssbarnea:
It would not work in docs. For pypi luckly they display number of open issues on the left sidebar so we could remove it.
Mainly I am ok to swap issues with codecov, seams as acceptable trade-off.

BeyondEvil (commented):
I'm fine with replacing open issues with codecov. 👍

gnikonorov:
Just removed the open issues badge, @ssbarnea and @BeyondEvil

ssbarnea (approved):

Issue #376, opened 2020-11-20
Is it possible to change the duration format from just seconds to something like hh:mm:ss?


brettnolan:
Is there a way to format the duration time to something like:
HH:MM:SS
or 
h hours, m mins, s secs 

I tried doing it within here, by changing the attribute value

```
@pytest.hookimpl(hookwrapper=True)
def pytest_runtest_makereport(item, call):
    outcome = yield
    report = outcome.get_result()
    report.description = str(item.function.__doc__)

    # change duration format here
    duration = getattr(report, "duration", 0.0)                          
    converted_duration = timedelta(seconds=duration)
    setattr(report, "duration", converted_duration)

```
However, **plugin.py** didn't like the format

```
INTERNALERROR>   File python-virtual-environments/python3_env/lib/python3.6/site-packages/pytest_html/plugin.py", line 162, in __init__
INTERNALERROR>     html.td(f"{self.time:.2f}", class_="col-duration"),
INTERNALERROR> TypeError: unsupported format string passed to datetime.timedelta.__format__
```

gnikonorov:
Hi @brettnolan let me look into this for you

gnikonorov:
@brettnolan can you paste the output of `pip list` please?

gnikonorov:
`duration` comes from the [pytest api](https://docs.pytest.org/en/stable/reference.html#pytest.reports.TestReport.duration) and must always be in seconds which is why you are getting the error you are seeing.

However, I think it would make sense for pytest to do the formatting on your behalf ( I don't really see a reason why 3721 is clear than 01:02:01, for example )

brettnolan:
Thanks!

Here is the 'pip list' output:


Package              Version
-------------------- ---------
apipkg               1.5
attrs                20.2.0
certifi              2020.6.20
click                7.1.2
click-completion     0.5.2
colorama             0.4.3
coverage             5.3
crayons              0.4.0
cycler               0.10.0
decorator            4.4.2
execnet              1.7.1
imageio              2.9.0
importlib-metadata   1.7.0
iniconfig            1.0.1
Jinja2               2.11.2
keyboard             0.13.5
kiwisolver           1.2.0
MarkupSafe           1.1.1
matplotlib           3.3.2
more-itertools       8.5.0
networkx             2.5
numpy                1.19.2
opencv-python        4.4.0.42
packaging            20.4
Pillow               7.2.0
pip                  20.2.3
pkg-resources        0.0.0
[...]

PR #380, opened 2020-11-26
Allow for report duration formatting

gnikonorov:
Allow for formatting the `Duration` column on the report. The supported format specifiers are those defined for [time.strftime](https://docs.python.org/3/library/time.html#time.strftime) plus `%f` ( for milliseconds ).

Closes https://github.com/pytest-dev/pytest-html/issues/376

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[coverage:report]
+precision = 2
+show_missing = true
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,63 @@
+{
+    "env": {
+        "browser": true,
+        "es2021": true
+    },
+    "extends": "eslint:recommended",
+    "parserOptions": {
+        "ecmaVersion": 12
+    },
+    "rules": {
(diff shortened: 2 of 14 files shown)


ssbarnea:
@gnikonorov Please disable the codecov comment, is noise/spam. We have an action check for it, if it fails we will be able to read it.

ssbarnea:
Are you sure we do not loose the sorting ability with it?

BeyondEvil (approved):
Nice! 👍

gnikonorov:
> @gnikonorov Please disable the codecov comment, is noise/spam. We have an action check for it, if it fails we will be able to read it.

Sure. I'll disable it when I'm at my computer later today.

gnikonorov:
> Are you sure we do not loose the sorting ability with it?

@ssbarnea I tested sorting the column with a few different 'duration_formatter' strings and with the default format and it worked

ssbarnea (approved):
I tested it locally and it seams ok.

Issue #374, opened 2020-11-20
Extras from reruns are attached to final test


VladimirPodolyan:
Deps:
- pytest==6.1.2
- pytest-rerunfailures==9.1.1
- pytest-selenium==2.0.1
- pytest-html==2.1.1/3.0.0

----

After updating `pytest-html` to `3.0.0` version in project requirements I got unexpected extras from `test reruns` in html report. Actually all extras from test reruns (logs/links/duration/screenshots) also applied to final test with failed/passed result.

### Ver 3.0.0
<img width="2266" alt="SM" src="https://user-images.githubusercontent.com/36446855/99747241-4ba94e00-2aeb-11eb-9fd6-e207f7ca2b8e.png">

### Ver 2.1.1
![SM1](https://user-images.githubusercontent.com/36446855/99748239-359c8d00-2aed-11eb-8102-bdfa59fa0514.png)

Ver 2.1.1 had a bug with missing screenshots, but other data correct. So my opinion that something was broken since ver 3.0.0

gnikonorov:
Hey @VladimirPodolyan thanks for raising this issue! Let me try and make a reproducer on my end.

gnikonorov:
For now, I'm thinking that every rerun should be its own entry in the table

gnikonorov:
![Screen Shot 2020-11-28 at 10 38 31 AM](https://user-images.githubusercontent.com/13026015/100519460-0083e080-3166-11eb-8294-ba0d0b703e60.png)

It looks like we already have one entry per rerun, but the data is populating incorrectly

gnikonorov:
I should have a PR up for this today or tomorrow, @VladimirPodolyan. Thank you for reporting this

VladimirPodolyan:
@gnikonorov Thank for respond and fixing the issue!

gnikonorov:
> @gnikonorov Thank for respond and fixing the issue! 

Of course @VladimirPodolyan ! I'll merge this in EOD today so it'll be available off of head and I'm planning to release to PyPI by the end of 12/2.

gnikonorov:
Hi @VladimirPodolyan you can get this functionality by installing the latest version of pytest-html from [PyPI](https://pypi.org/project/pytest-html/)

VladimirPodolyan:
@gnikonorov The report looks fine now. Thanks for the poke!

gnikonorov:
> @gnikonorov The report looks fine now. Thanks for the poke!

[...]

PR #387, opened 2020-11-29
Stop attaching test reruns to final test report entries

gnikonorov:
Stop attaching the rerun entries created by [pytest-rerunfailures](https://github.com/pytest-dev/pytest-rerunfailures) to  reports.

When we introduced our post processing step we incorrectly lumped in rerunfailures reruns ( see https://github.com/pytest-dev/pytest-html/pull/271 ), which caused this issue. To fix it, we now treat each rerun as its own test which is what the original behavior was. I confirmed that behavior is the same as v2.1.1 now.

I'd also like to make a release after this merges, since this is a bugfix for the `v3.0.0` release if no one objects

Closes https://github.com/pytest-dev/pytest-html/issues/374

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,10 @@ Release Notes
 
 **3.1.0 (unreleased)**
 
+* Stop attaching test reruns to final test report entries (`#374 <https://github.com/pytest-dev/pytest-html/issues/374>`_)
+
+  * Thanks to `@VladimirPodolyan <https://github.com/VladimirPodolyan>`_ for reporting and `@gnikonorov <https://github.com/gnikonorov>`_ for the fix
+
 * Allow for report duration formatting (`#376 <https://github.com/pytest-dev/pytest-html/issues/376>`_)
 
   * Thanks to `@brettnolan <https://github.com/brettnolan>`_ for reporting and `@gnikonorov <https://github.com/gnikonorov>`_ for the fix
--- a/src/pytest_html/plugin.py
+++ b/src/pytest_html/plugin.py
@@ -690,7 +690,7 @@ class HTMLReport:
             #    extras, and when it swapped from pass
             for test_report in test_reports:
                 if test_report.outcome == "rerun":
-                    # reruns are seperate test runs for all intensive purposes
+                    # reruns are separate test runs for all intensive purposes
                     self.append_rerun(test_report)
                 else:
                     full_text += test_report.longreprtext
--- a/src/pytest_html/plugin.py
+++ b/src/pytest_html/plugin.py
@@ -157,7 +157,7 @@ class HTMLReport:
             if getattr(report, "when", "call") != "call":
                 self.test_id = "::".join([report.nodeid, report.when])
             self.time = getattr(report, "duration", 0.0)
-            self.formatted_time = getattr(report, "formatted_duration", 0.0)
+            self.formatted_time = self._format_time(report)
             self.outcome = outcome
             self.additional_html = []
             self.links_html = []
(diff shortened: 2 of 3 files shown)


ssbarnea:
IMHO retries should be visible, they may include essential debugging information, in fact we should also make them visible in the UI with some other color than green.

BeyondEvil:
> IMHO retries should be visible, they may include essential debugging information, in fact we should also make them visible in the UI with some other color than green.

I agree. But I think it might be worth doing a bugfix release, and then adding the functionality you mention in a separate release.

gnikonorov:
@ssbarnea the retries are visible. The issue is that without this PR we take all the retry data and add it to the original run. With this PR the data for each retry row is not added to the original run. Does that make sense?

For example. Here is what a report off of v3.0.0 looks like:
Screen Shot 2020-11-29 at 10 08 13 AM

Here is the same test run used to generate the above report but off of this PR instead:
Screen Shot 2020-11-29 at 10 08 49 AM


You can see that the retry rows still exist and have the proper data. We aren't, however, adding everything into the final pass/fail run of the retry as we are off of head though which is a bug. This didn't happen in 2.1.1 and shouldn't happen IMO.

gnikonorov:
> > IMHO retries should be visible, they may include essential debugging information, in fact we should also make them visible in the UI with some other color than green.
> 
> I agree. But I think it might be worth doing a bugfix release, and then adding the functionality you mention in a separate release.

@BeyondEvil, that functionality already exists, please see the above comment

BeyondEvil:
> @BeyondEvil, that functionality already exists, please see the above comment

Yepp, you're right! 👍

BeyondEvil (commented):
Why is the time/duration stuff in this PR?

gnikonorov:
> Why is the time/duration stuff in this PR?

I moved time formatting to the `TestResult` class since it makes no sense to format on the report as we merge reports for test results. Formatting should be done only once when we compute the test result to show to the user. The way we were doing it was that we formatted once per report, and the later merged reports which is wrong as we first need to accumulate all the report times and only then format them for display

gnikonorov:
> > Why is the time/duration stuff in this PR?
> 
> I moved time formatting to the `TestResult` class since it makes no sense to format on the report as we merge reports for test results. Formatting should be done only once when we compute the test result to show to the user. The way we were doing it was that we formatted once per report, and the later merged reports which is wrong as we first need to accumulate all the report times and only then format them for display

And seeing that we needed to touch the logic anyway as part of this change ( line 718 in the original version of plugin.py ), I figured we may as well move it to the proper place instead of moving it but keep a known bug that we'd have to fix for an unreleased feature later on

BeyondEvil:
> > > Why is the time/duration stuff in this PR?
> > 
> > 
> > I moved time formatting to the `TestResult` class since it makes no sense to format on the report as we merge reports for test results. Formatting should be done only once when we compute the test result to show to the user. The way we were doing it was that we formatted once per report, and the later merged reports which is wrong as we first need to accumulate all the report times and only then format them for display
> 
> And seeing that we needed to touch the logic anyway as part of this change ( line 718 in the original version of plugin.py ), I figured we may as well move it to the proper place instead of moving it but keeping a known bug that we'd have to fix for an unreleased feature later on

Would it makes sense to have that as a separate PR? Since it's (sort of) unrelated to reruns? 🤔

gnikonorov:
> > > > Why is the time/duration stuff in this PR?
> > > 
> > > 
> > > I moved time formatting to the `TestResult` class since it makes no sense to format on the report as we merge reports for test results. Formatting should be done only once when we compute the test result to show to the user. The way we were doing it was that we formatted once per report, and the later merged reports which is wrong as we first need to accumulate all the report times and only then format them for display
> > 
> > 
> > And seeing that we needed to touch the logic anyway as part of this change ( line 718 in the original version of plugin.py ), I figured we may as well move it to the proper place instead of moving it but keeping a known bug that we'd have to fix for an unreleased feature later on
> 
> Would it makes sense to have that as a separate PR? Since it's (sort of) unrelated to reruns? 🤔

We could, but I think that would make this change dirty since it would look like this:
```
                 if test_report.outcome == "rerun":
                     # reruns are separate test runs for all intensive purposes
+                    test_report.duration = getattr(test_report, "duration", 0.0)
+                    test_report.formatted_duration = self._format_duration(test_report)
                     self.append_rerun(test_report)
```

And we would also be doing it immediately below on https://github.com/pytest-dev/pytest-html/blob/master/src/pytest_html/plugin.py#L717 which is confusing to read.

I'd argue this is related as this PR is really fixing all postprocessing related to dynamic reruns, and this move of the formatting to `TestResult` is a direct result of that since if we didn't have to change the postprocessing logic the duration formatting on line 718 would make sense.

It seems like a big PR but all I did was move a function from one place to another and it's call from one place to another so it's 2 lines. The method is the same 

@BeyondEvil

BeyondEvil:
> It seems like a big PR but all I did was move a function from one place to another and it's call from one place to another so it's 2 lines. The method is the same
> 
> @BeyondEvil

Fair enough. 😊

BeyondEvil (approved):
Send it! 🚀

gnikonorov:
Thanks @BeyondEvil! I'll merge in tomorrow assuming @ssbarnea has no issues with it.

Then I'll release `v3.1.0` on 12/2

Issue #388, opened 2020-11-29
[BUG] Not all missing CSS files are reported


prakhargurunani:
## What's the error
Assume none of the CSS files are present in the directory.
if I run `pytest --html=report.html --css=abc.css --css=xyz.css`, It should give me error stating that `abc.css` and `xyz.css` files were not found.

## Actual Output
But the current pytest version gives error for only the first CSS file `abc.css` and ignores `xyz.css` (which means it will report error for only the first `--css` argument if it's file is not found).
```bash
$ pytest --html=report.html --css=abc.css --css=xyz.css

    INTERNALERROR> Traceback (most recent call last):
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/_pytest/main.py", line 253, in wrap_session
    INTERNALERROR>     config._do_configure()
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/_pytest/config/__init__.py", line 982, in _do_configure
    INTERNALERROR>     self.hook.pytest_configure.call_historic(kwargs=dict(config=self))
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/pluggy/hooks.py", line 308, in call_historic
    INTERNALERROR>     res = self._hookexec(self, self.get_hookimpls(), kwargs)
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/pluggy/manager.py", line 93, in _hookexec
    INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/pluggy/manager.py", line 84, in <lambda>
    INTERNALERROR>     self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/pluggy/callers.py", line 208, in _multicall
    INTERNALERROR>     return outcome.get_result()
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/pluggy/callers.py", line 80, in get_result
    INTERNALERROR>     raise ex[1].with_traceback(ex[2])
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/pluggy/callers.py", line 187, in _multicall
    INTERNALERROR>     res = hook_impl.function(*args)
    INTERNALERROR>   File "/usr/local/lib/python3.9/site-packages/pytest_html/plugin.py", line 92, in pytest_configure
[...]

PR #390, opened 2020-11-29
Setup reporting for missing CSS file

prakhargurunani:
fixes: #388 

Returns missing CSS files in `OSError` according to the order they were passed.


gnikonorov (changes requested):
Thanks for raising this @prakhargurunani!

Can you please also add:
- A [changelog entry](https://github.com/pytest-dev/pytest-html/blob/master/CHANGES.rst)
- Some [tests for this change](https://github.com/pytest-dev/pytest-html/blob/master/testing/test_pytest_html.py)

prakhargurunani:
@gnikonorov I have only changed the way of reporting the missing CSS files. I can't see any new test cases for this change. If any, Please let me know.

gnikonorov:
> @gnikonorov I have only changed the way of reporting the missing CSS files. I can't see any new test cases for this change. If any, Please let me know.

In the [run](https://github.com/pytest-dev/pytest-html/blob/3f6368693f4316372c0b7113ecf0ccd1300e7a05/testing/test_pytest_html.py#L39) method you can pass a variable list of arguments via `*args`. You could pass a list of css files that don't exist and confirm behavior via the [css flag](https://github.com/pytest-dev/pytest-html/blob/3f6368693f4316372c0b7113ecf0ccd1300e7a05/src/pytest_html/plugin.py#L66)

prakhargurunani:
Thanks @gnikonorov !

src/pytest_html/plugin.py
```diff
@@ -87,9 +87,17 @@ def pytest_addoption(parser):
...
+        files_not_in_path = []
+        for filenos in range(len(config.getoption("css"))):
+            csspath = config.getoption("css")[filenos]
+            if filenos <= len(config.getoption("css")):
+                if not os.path.exists(csspath):
+                    files_not_in_path.append(csspath)
+
+        if len(files_not_in_path) != 0:
+            no_files_found = ", ".join(str(x) for x in files_not_in_path)
+            raise OSError(f"No such file or directory: '{no_files_found}'")
```
BeyondEvil, line 99:
```suggestion
        missing_css_files = []
        for csspath in config.getoption("css"):
            if not os.path.exists(csspath):
                missing_css_files.append(csspath)

        if missing_css_files:
            raise IOError(
                f"Missing CSS file{'s' if len(missing_css_files) > 1 else ''}: '{', '.join(missing_css_files)}'"
            )
```

BeyondEvil (changes requested):

src/pytest_html/plugin.py
prakhargurunani: Thanks @BeyondEvil !

prakhargurunani:
> In the [run](https://github.com/pytest-dev/pytest-html/blob/3f6368693f4316372c0b7113ecf0ccd1300e7a05/testing/test_pytest_html.py#L39) method you can pass a variable list of arguments via `*args`. You could pass a list of css files that don't exist and confirm behavior via the [css flag](https://github.com/pytest-dev/pytest-html/blob/3f6368693f4316372c0b7113ecf0ccd1300e7a05/src/pytest_html/plugin.py#L66)

@gnikonorov Done.

BeyondEvil (dismissed):
Thanks for this @prakhargurunani !

I'm satisfied but I'll let @gnikonorov have the last word (and/or merge). 👍

testing/test_pytest_html.py
```diff
@@ -1010,7 +1010,15 @@ def test_css_invalid(self, testdir, recwarn):
         result = testdir.runpytest("--html", "report.html", "--css", "style.css")
         assert result.ret
         assert len(recwarn) == 0
-        assert "No such file or directory: 'style.css'" in result.stderr.str()
+        assert "Missing CSS file: style.css" in result.stderr.str()
+
+    def test_multiple_css_files(self, testdir, recwarn):
```
gnikonorov, line 1015: Can you add a test for where only part of the passed files are missing? You can use https://github.com/pytest-dev/pytest-html/blob/master/testing/test_pytest_html.py#L321 as an example for creating adhoc files in tests.

Also, please use [@pytest.mark.parametrize](https://docs.pytest.org/en/stable/parametrize.html) to combine your tests with the test above. It'll be a lot cleaner than having 2-3 separate tests for the same feature ( reporting of missing css files )

[Here](https://github.com/pytest-dev/pytest-html/blob/master/testing/test_pytest_html.py#L482) is an example of how we parametrize in this file

gnikonorov (changes requested):
One last comment and then I think this is good to merge @prakhargurunani

testing/test_pytest_html.py
prakhargurunani: @gnikonorov See https://github.com/pytest-dev/pytest-html/pull/390/commits/8ccc1045cfa80b463994feb5f418f96a613bcfd9
Have I used `@pytest.mark.parametrize` correctly ?

testing/test_pytest_html.py
```diff
@@ -1005,12 +1005,31 @@ def test_css(self, testdir, recwarn, colors):
...
         testdir.makepyfile("def test_pass(): pass")
-        result = testdir.runpytest("--html", "report.html", "--css", "style.css")
+        result = testdir.runpytest(
+            "--html", "report.html", "--css", file1, "--css", file2
+        )
+        assert result.ret
+        assert len(recwarn) == 0
+        assert expected in result.stderr.str()
+
+    def test_some_css_files(self, testdir, recwarn):
```
gnikonorov, line 1024: You can merge this test with the test above, just create the valid css file in the body of the test.

testing/test_pytest_html.py
```diff
@@ -1005,12 +1005,31 @@ def test_css(self, testdir, recwarn, colors):
...
 
-    def test_css_invalid(self, testdir, recwarn):
+    @pytest.mark.parametrize(
+        "file1,file2,expected",
+        [
+            ("style.css", "", "Missing CSS files: style.css,"),
+            ("abc.css", "xyz.css", "Missing CSS files: abc.css, xyz.css"),
+        ],
+    )
+    def test_css_invalid(self, testdir, recwarn, file1, file2, expected):
```
gnikonorov, line 1015: This is correct, but you could make it even more clean by parametrizing with a list of files instead of passing `file1` and `file2`. For example:
```
$ cat test_parametrize.py
import pytest

@pytest.mark.parametrize("arg", ["arg1", "arg2", ["arg3.a", "arg3.b"]])
def test_list(arg):
     assert arg is not None
$
```

In the example above the test is run 3 times. First with `"arg1"`, then `"arg2"`, and finally with the list `["arg3.a", "arg3.b"]`

This way your test can be expanded on in the future without changing its signature every time, and it becomes easier to reason about

gnikonorov (changes requested):

testing/test_pytest_html.py
gnikonorov: Almost @prakhargurunani! Please see my last review

gnikonorov:
I'll take a look tonight @prakhargurunani! Sorry, I've been busy

gnikonorov (changes requested):
Please fix the changelog. After that I'm fine with approving

testing/test_pytest_html.py
```diff
@@ -1005,12 +1005,32 @@ def test_css(self, testdir, recwarn, colors):
...
+    @pytest.mark.parametrize(
+        "files",
+        [
+            "style.css",
+            ["abc.css", "xyz.css"],
+            "testdir.makefile('.css', * {color: 'white'}",
+        ],
+    )
+    def test_css_invalid(self, testdir, recwarn, files):
+        assert files is not None
```
gnikonorov, line 1017: Is this debug code? `files` should never be `None` based on your parametrization

testing/test_pytest_html.py
```diff
@@ -1005,12 +1005,32 @@ def test_css(self, testdir, recwarn, colors):
...
+            result = testdir.runpytest(
+                "--html", "report.html", "--css", file1, "--css", file2
+            )
+        else:
+            result = testdir.runpytest("--html", "report.html", "--css", path)
         assert result.ret
         assert len(recwarn) == 0
-        assert "No such file or directory: 'style.css'" in result.stderr.str()
+        if isinstance(files, list):
+            assert files[0], files[1] in result.stderr.str()
```
gnikonorov, line 1031: Sorry I missed this, but this doesn't do what you expect. You aren't actually checking for `files[0]`. This just checks whether or not `files[0]` is truthy and executes `files[1] in result.stderr.str()` if it's not. For example:
```
>>> l = [1,2,3]
>>> assert 0, 2 in l
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError: True
>>> assert 44, 2 in l
```

gnikonorov (changes requested):

gnikonorov (approved):
LGTM, thanks @prakhargurunani !

I'll leave the final review/merge up to @BeyondEvil

BeyondEvil (dismissed):
Thanks for all the hard work @prakhargurunani !

This looks great!

Can you squash and rebase so we can merge?

BeyondEvil:
@gnikonorov docs are failing, not sure what to do here? 🤔

gnikonorov:
@BeyondEvil docs are failing because the RTD build was enabled before the PR for RTD was merged. Once https://github.com/pytest-dev/pytest-html/pull/402 goes in the issue will go away

BeyondEvil (approved):
Thanks everyone!

PR #423, opened 2020-12-15
Move the changelog to read the docs

gnikonorov:
Move the changelog to read the docs, since it would be a better user experience to have it with the rest of the documentation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
       - id: rst
         name: rst
         entry: rst-lint --encoding utf-8
-        files: ^(CHANGES.rst|development.rst|README.rst)$
+        files: ^(README.rst)$
         language: python
         additional_dependencies: [pygments, restructuredtext_lint]
   - repo: https://github.com/elidupuis/mirrors-sass-lint
--- a/README.rst
+++ b/README.rst
@@ -26,7 +26,7 @@ Resources
 ---------
 
 - `Documentation <https://pytest-html.readthedocs.io/en/latest/>`_
-- `Release Notes <http://github.com/pytest-dev/pytest-html/blob/master/CHANGES.rst>`_
+- `Release Notes <https://pytest-html.readthedocs.io/en/latest/changelog.html>`_
 - `Issue Tracker <http://github.com/pytest-dev/pytest-html/issues>`_
 - `Code <http://github.com/pytest-dev/pytest-html/>`_
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,19 +1,27 @@
-Release Notes
--------------
+Changelog
+=========
 
-**3.2.0 (unreleased)**
+Versions follow `Semantic Versioning`_ (``<major>.<minor>.<patch>``).
+
+Version History
+---------------
+
+3.2.0 (unreleased)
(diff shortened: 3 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -49,7 +49,7 @@ repos:
       - id: rst
         name: rst
         entry: rst-lint --encoding utf-8
-        files: ^(CHANGES.rst|development.rst|README.rst)$
+        files: ^(README.rst)$
```
BeyondEvil, line 52: Don't we want to lint the removed rst-files?

gnikonorov: > Don't we want to lint the removed rst-files?

We do, and RTD does lint for us. This lint is for files under the project root though and so should be removed since we're asking for it to lint file that no longer exist

BeyondEvil: Got it, thanks! 👍 

Can we still lint on commit somehow? So that you don't have to wait for RTD to do it.

gnikonorov: I'd need to double check @BeyondEvil ( I think so ) but regardless of whether we do or not this line achieves nothing and should be removed since the files no longer exist.

I made a note to double check linting

gnikonorov: We also have a tox command to build documentation and it runs in GHA on all PRs, so we definitely don't need to wait on upload to RTD for feedback @BeyondEvil

BeyondEvil: Right, sphinx, so can I assume sphinx will catch linting issues then? @gnikonorov

gnikonorov: > Right, sphinx, so can I assume sphinx will catch linting issues then? @gnikonorov 
> 
> 

I think so. I made a note to double check this. 

We also can't use this linter since the RST syntax is different ( I remember trying to use it and having it fail in the original doc migration )

ssbarnea:
How about moving the changelog outside SCM to https://github.com/pytest-dev/pytest-html/releases

This would make much easier to maintain and we would only have to manually copy/paste old entries and add them to exiting tags. After that, we only remove the changelog file and update the dos to point to github release page.

After that, we no longer need to remind to update the changelog on changes, the bot does it for us.

gnikonorov:
Doesn't that still need more work @ssbarnea ? Or is it good to go

gnikonorov:
> Doesn't that still need more work @ssbarnea ? Or is it good to go

For example it's still pointing at v3.0.0 but we're at 3.1.1 currently.

I don't mind moving to it when it's set up, this is just for now to be consistent since all our docs are now in RTD

.pre-commit-config.yaml
gnikonorov: Opened an issue to investigate linting RTD @BeyondEvil (https://github.com/pytest-dev/pytest-html/issues/428). Can you please review this change? I'd like to get it merged if there are no objections on your end. I'll tackle any linting requirements ( if needed ) in a follow up PR

gnikonorov:
Bump for review, @BeyondEvil / @ssbarnea

BeyondEvil (approved):
LGTM!

gnikonorov:
Merging this in with the understanding that the changelog will be deprecated when the release drafter is properly set up

Issue #418, opened 2020-12-13
Remove phantomjs dependency


BeyondEvil:
Since we're using chrome headless, we can remove the phantomjs dependency from package.js.

PR #424, opened 2020-12-15
remove phantomjs dependency

gnikonorov:
Removes the phantomjs dependency as per https://github.com/pytest-dev/pytest-html/issues/418

Closes https://github.com/pytest-dev/pytest-html/issues/418


BeyondEvil (approved):
Thanks! 👍

PR #425, opened 2020-12-15
properly classify all npm dependencies

gnikonorov:
Stop classifying dev dependencies as non dev dependencies in the `package.json` file


BeyondEvil (approved):
👍

PR #426, opened 2020-12-16
Initial backend for Next Gen

BeyondEvil:
The initial backend.

It will create a JSON file with the title, environment (from metadata) and all tests.

Any calculations, summary of passed/failed, durations etc. will be left to the frontend.

`extras` and `hooks` are not implemented yet.

### DIFF
--- a/src/pytest_html/plugin.py
+++ b/src/pytest_html/plugin.py
@@ -19,6 +19,7 @@ from os.path import isfile
 
 import pytest
 from _pytest.logging import _remove_ansi_escape_sequences
+from _pytest.pathlib import Path
 from py.xml import html
 from py.xml import raw
 
@@ -101,7 +102,9 @@ def pytest_configure(config):
         if not hasattr(config, "workerinput"):
             # prevent opening htmlpath on worker nodes (xdist)
             config._html = HTMLReport(htmlpath, config)
+            config._next_gen = NextGenReport(config, Path("nextgendata.json"))
             config.pluginmanager.register(config._html)
+            config.pluginmanager.register(config._next_gen)
 
 
 def pytest_unconfigure(config):
@@ -109,6 +112,11 @@ def pytest_unconfigure(config):
     if html:
         del config._html
         config.pluginmanager.unregister(html)
+
+    next_gen = getattr(config, "_next_gen", None)
+    if next_gen:
+        del config._next_gen
+        config.pluginmanager.unregister(next_gen)
 
 
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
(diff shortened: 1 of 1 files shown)


gnikonorov:
🎉 Thanks @BeyondEvil ! I'll take a look as soon as I can

src/pytest_html/plugin.py
```diff
@@ -757,3 +767,64 @@ def pytest_sessionfinish(self, session):
 
     def pytest_terminal_summary(self, terminalreporter):
         terminalreporter.write_sep("-", f"generated html file: file://{self.logfile}")
+
+
+class NextGenReport:
```
gnikonorov, line 772: can you move this into a separate file? I want to break apart `plugin.py` so it's easier to read/develop it ( the PR for that is [here](https://github.com/pytest-dev/pytest-html/pull/427) )

BeyondEvil: Good idea!

Can you create a ticket for it in the project? Or I will forget.

gnikonorov: Sure @BeyondEvil. Added a card in the todo column

src/pytest_html/nextgen.py
```diff
@@ -0,0 +1,66 @@
+import json
+from typing import Any
```
gnikonorov, line 2: Thanks for adding typing! Can you add [mypy](https://mypy.readthedocs.io/en/stable/) to the pre commit config as part of the next gen migration so we can enforce it?

We could just follow what [pytest](https://github.com/pytest-dev/pytest/blob/master/.pre-commit-config.yaml#L54) does. I think it's the easiest way to do it.

I don't think it's necessary in this PR, but can we do it as part of next gen? It would be good to have 100% typing if possible

src/pytest_html/nextgen.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _write(self):
+        try:
+            data = json.dumps(self._data)
+        except TypeError:
+            data = cleanup_unserializable(self._data)
+            data = json.dumps(data)
+
+        with self._data_file.open("w", buffering=1, encoding="UTF-8") as f:
+            f.write("const jsonData = ")
```
gnikonorov, line 31: can we combine these `f.write` calls on an f string line `f"const jsonData = {data}\n"`?

src/pytest_html/nextgen.py
```diff
@@ -0,0 +1,66 @@
...
+            f.write(data)
+            f.write("\n")
+
+    @pytest.hookimpl(trylast=True)
+    def pytest_sessionstart(self, session):
```
gnikonorov, line 36: Wouldn't all the `self._write` calls overwrite `self._data_file`?

gnikonorov (commented):
Other than these comments, LGTM! Thank you again @BeyondEvil !

src/pytest_html/nextgen.py
gnikonorov: I opened an issue to track typing (https://github.com/pytest-dev/pytest-html/issues/434)

I also added mypy to our linting target and began work on typing the project in https://github.com/pytest-dev/pytest-html/pull/435

Issue #429, opened 2020-12-18
CI pipeline is broken with error 'Error: IS_WINDOWS is not defined'


gnikonorov:
This seems to be related to https://github.com/actions/setup-python/issues/171. It's currently affecting the master build and PRs.

FYI @BeyondEvil / @ssbarnea

gnikonorov:
https://github.com/pytest-dev/pytest-html/pull/430 rolls back to the last stable version so we can unbreak the CI pipeline

hugovk:
Will be fixed by https://github.com/actions/setup-python/pull/172.

gnikonorov:
Thanks for the news, @hugovk! Once the fix goes to master when will a new release be made?

gnikonorov:
This could already be fixed as per https://github.com/actions/setup-python/issues/171#issuecomment-748114163. I'll confirm tonight (EST)

hugovk:
Yep, looks fixed on my repo: https://github.com/hugovk/tinytext/actions/runs/428941900

gnikonorov:
Can confirm it's fixed on PRs and master in this repo as well. Closing this issue then since no changes are needed

PR #430, opened 2020-12-18
temporarily use last stable version of setup-python until error in latest is resolved

gnikonorov:
This PR reverts `setup-python` to the last stable version until https://github.com/actions/setup-python/issues/171 is resolved. Without this, our CI pipelines are broken since setup-python always fails

The issue tracking this is https://github.com/pytest-dev/pytest-html/issues/429. I do not want to close it until https://github.com/actions/setup-python/issues/171 is resolved as we still need to pin to latest after that happens.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -22,7 +22,8 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - name: Set up Python
-      uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


gnikonorov:
Please do not merge this if you approve. This may already be fixed as per https://github.com/actions/setup-python/issues/171#issuecomment-748114163

I can confirm tonight (EST)

gnikonorov:
Change is no longer needed as per https://github.com/pytest-dev/pytest-html/issues/429#issuecomment-748356732

Closing

Issue #399, opened 2020-12-03
Control visibility of test results sections by query param


TheCorp:
Just like we can control collapsed test categories via query param (https://github.com/pytest-dev/pytest-html#display-options) would it also make sense to control whether test results sections are hidden or not (checkbox checked or not effectively) by query param? Something like:

`?checked=Passed,XFailed,Skipped`

The reason I am asking this is that it seems like this is nearly impossible to control before the page loads. We'd have to inject some JS that calls the "filter_tableI()" method for each section we want to check/uncheck or modify the <tbody> to have the "hidden" attribute before the page is generated.

I understand we can wholesale not render elements for any section using the output hooks (https://github.com/pytest-dev/pytest-html#modifying-the-results-table) but that would remove the ability for anyone to see the results if they were interested in seeing them. We have test output with hundreds of passed/skipped outputs usually and only a few failed ones. It would be nice if all of those were hidden by default but still viewable if the user wanted to view them. 

PR that did the original feature: https://github.com/pytest-dev/pytest-html/pull/168

BeyondEvil:
Feel free to contribute a PR with a proposed solution! @TheCorp 😊 👍

TheCorp:
Awesome. I just wanted to make sure the idea itself didn't sound stupid 😄

BeyondEvil:
Definitely not! @TheCorp

gnikonorov:
Hey @TheCorp I'm going to implement this. Let me know if you're already working on a solution. If not, I'll make a PR in the coming days

TheCorp:
Oh wow @gnikonorov thank you! No I haven't had the chance to start on this yet so go for it. Really appreciate you taking this on.

gnikonorov:
Of course @TheCorp. I'll have something up today or tomorrow

Issue #432, opened 2020-12-20
Migrate JavaScript test framework to more popular alternative


gnikonorov:
We currently use [QUnit](https://qunitjs.com/) for unit testing our JavaScript.

I propose that we move to a more modern testing framework ( the top five listed [here](https://codeburst.io/top-5-javascript-testing-frameworks-to-choose-out-of-in-2020-6903ae94461e) seem to pop up a bit ).

I would like to add that any library we migrate to should have good support for testing URL parameters. `QUnit` forces the definition of one test script per query string, which I think isn't ideal, since it's pretty cumbersome: https://stackoverflow.com/questions/27388596/pass-url-parameter-to-qunit-test-using-grunt

@BeyondEvil maybe this is something we could tackle as part of the next gen migration?

BeyondEvil:
> @BeyondEvil maybe this is something we could tackle as part of the next gen migration?

Yes!

PR #433, opened 2020-12-20
Implement the visible URL query parameter to control visibility of test results on page load.

gnikonorov:
Implement a new query parameter, `visible`, to control visibility of test results on page load.

Note that `QUnit` makes testing this cumbersome. I tested this locally with all possible combinations and things worked as expected ( as did not passing this argument ). I opened https://github.com/pytest-dev/pytest-html/issues/432 to migrate off QUnit and noted the need for URL parameter tests in the issue.

Closes https://github.com/pytest-dev/pytest-html/issues/399

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,10 @@ Version History
 3.2.0 (unreleased)
 ~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


BeyondEvil (approved):
LGTM! 👍

TheCorp:
Beautiful! Thanks!

PR #445, opened 2021-02-16
Add logo to htmlreport.

nirajnagle:
Enhancement to add logo to Test Reports.
Screen Shot 2021-02-15 at 11 11 59 PM

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,23 @@ Versions follow `Semantic Versioning`_ (``<major>.<minor>.<patch>``).
 
 Version History
 ---------------
+3.2.1(2021-02-15)
+
+* Implement functionalitly to add/customize logo on Tests reports page.
+
+  * Report Logo
+
+  * By default the report logo will be None.
+
+  * you can edit it by using the pytest_html_report_logo hook
+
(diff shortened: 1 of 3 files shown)


nirajnagle:
Hello
@BeyondEvil @gnikonorov 
Bumping this up for review.
Thanks

gnikonorov:
Hi @nirajnagle I've been busy with work, but perhaps @BeyondEvil or @ssbarnea can review

nirajnagle:
Thanks @BeyondEvil

giuliohome:
I'm also looking to add a logo and I found this:
@BeyondEvil any chances it can be merged now or has been found another solution in the meantime? 
Or maybe should I consider to fork and apply it on my own?
Thank you again.

giuliohome:
BTW I've also tested this patch but it looks like the call of the hook is missing in the report generator: I had to add the hool call in the code of html_report.py to make it work.

However I've already solved my requirement via css, as explained above (it appears better to me)

BeyondEvil:
I wonder if the easiest and best solution here is to just allow HTML in the `title` hook. That way the user can add a logo using  HTML and CSS (via the `--css` option).

PR #447, opened 2021-03-16
use --show-capture as filter

swbchangle:
use --show-capture the same way as in _pytest.terminal module.

This is related to issue #171. `--show-capture=no|stdout|stderr|log|all` can be used to control which captured log goes into html file. `--show-capture=no` disables all captured logs. For failed tests, longrepr is always included in html, even if `--show-capture=no`


BeyondEvil:
I believe v4 now behaves as expected.

PR #454, opened 2021-05-07
Support attaching arbitrary binary files

eqvinox:
(Our testcases produce wireshark pcap files...)

### DIFF
--- a/src/pytest_html/extras.py
+++ b/src/pytest_html/extras.py
@@ -8,6 +8,7 @@ FORMAT_JSON = "json"
 FORMAT_TEXT = "text"
 FORMAT_URL = "url"
 FORMAT_VIDEO = "video"
+FORMAT_BINARY = "binary"
 
 
 def extra(content, format_type, name=None, mime_type=None, extension=None):
--- a/src/pytest_html/result.py
+++ b/src/pytest_html/result.py
@@ -116,6 +116,15 @@ class TestResult:
                     content, extra_index, test_index, extra.get("extension")
                 )
 
+        elif extra.get("format_type") == extras.FORMAT_BINARY:
+            content = extra.get("content")
+            if self.self_contained:
+                href = self._data_uri(content, mime_type=extra.get("mime_type"))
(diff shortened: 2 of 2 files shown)


BeyondEvil:
Does it really makes sense to add this in a self-contained report?

eqvinox:
> Does it really makes sense to add this in a self-contained report?

For any network-heavy software being tested, pcap files are equivalent (if not above) in importance to logfiles, so I don't really have any doubt in a "yes" to that. Especially since that's just one (pretty wide) use case, there's probably a bunch of others. Trying to think of some others, testing (de-)serialization feels like something where attaching random binaries would be pretty useful too.

BeyondEvil:
How would you consume that pcap file from the self-contained report? It's going to be base64 encoded.

But maybe I'm misunderstanding something here? @eqvinox

eqvinox:
> How would you consume that pcap file from the self-contained report? It's going to be base64 encoded.
> 
> But maybe I'm misunderstanding something here? @eqvinox

@BeyondEvil when you click the link for the file, the browser will open a file download dialog and save it decoded. (The download dialog would probably list wireshark too, as an application to directly open the file with. The file is first saved to `/tmp` for that, same way all "open with" downloads work.)

BeyondEvil:
Ah, got it. Ok, I'll add this to next-gen. Thanks!

BeyondEvil:
Missed this is during initial work of v4. Will try to get to it now.

BeyondEvil:
> > How would you consume that pcap file from the self-contained report? It's going to be base64 encoded.
> > But maybe I'm misunderstanding something here? @eqvinox
> 
> @BeyondEvil when you click the link for the file, the browser will open a file download dialog and save it decoded. (The download dialog would probably list wireshark too, as an application to directly open the file with. The file is first saved to `/tmp` for that, same way all "open with" downloads work.)

I'm not seeing the behavior described here.

When I click the link, it just opens a new tab displaying the path to the binary.

Issue #449, opened 2021-04-08
Duration column does not sort numerically


JCBdM:
When ordering the test list by duration I obtain:

| Result | Test   | Duration | Links |
| ------ | -------| ---------| ----- |
| Passed | test_a | 9.31     |       |
| Passed | test_d | 47.04    |       |
| Passed | test_b | 4.52     |       |
| Passed | test_c | 1.63     |       |

This order is done by alphanumeric and not by duration.

fenchu:
See https://github.com/pytest-dev/pytest-html/issues/413

SoCalLongboard:
If PR #464 solves the issue, that would be fantastic. Is anyone looking into its efficacy?

akhilramkee:
@BeyondEvil Did you get a chance to look into this feature? 

I can help with this, as I would like for this feature to work for my project's usecase.

BeyondEvil:
> ~@BeyondEvil Did you get a chance to look into this feature?~
> 
> ~I can help with this, as I would like for this feature to work for my project's usecase.~
> 
> [Ignore the above]
> 
> Thanks for taking care of this in [ae54ee6](https://github.com/pytest-dev/pytest-html/commit/ae54ee6277b0a6b4b3ee760bc09b8799046342e8).

Can you try 4.0.0rc5?

Issue #413, opened 2020-12-12
Numeric column sorting disappeared in v3


lukjak:
Hello,

I've noticed that in v3 the ability to sort columns in report by numeric value has been removed. ("Numeric" class is no more detected and relevant key sorting function is removed).
Can I restore this capability from v2? Or is there any reason it's not in v3 anymore?

gnikonorov:
Hey @lukjak thank you for filing this issue!. Could you attach a screenshot with what you're describing? It'll be easier to investigate

gnikonorov:
If I understand correctly, you're referring to 'numeric' class. That was removed in https://github.com/pytest-dev/pytest-html/commit/b57ca0bcf25e77bc95f1216eed51c7d235cb270e. The reason it was removed is because it was only used to sort values in the 'Duration' column. That column used to be reported as a double value of seconds+milliseconds but is now reported as a formatted time string ( see [this](https://pytest-html.readthedocs.io/en/latest/user_guide.html#display-options) link ).

All sorting works exactly as before for report columns with this change. Is there a bug you're seeing?

lukjak:
Hello

I observed it for a custom column (added "officially" via hooks) - so there is no issue in the default setup. It's used for test results ordering - so numbers are the most convenient.

fenchu:
Duration fails in v3x:

```
vvvDuration (highest first)
--
8.89
8.72
5.64
35.34
3.95
236.57
2.97
17.91
14.42

and the other way:


^^^Duration (lowest first)
--
14.42
17.91
2.97
236.57
3.95
35.34
5.64
8.72
8.89
```

fenchu:
You can patch this in main.js by adding a simple check for number and float and then subtract to find who is largest:

```
function numOrFloat(x) {
    let regexPattern = /^-?[0-9\.]+$/;
    if (regexPattern.test(x)) {
        return true;
    }
    return false;
}

function sort(items, keyFunc, reversed) {
    const sortArray = items.map(function(item, i) {
        return [keyFunc(item), i];
    });

    sortArray.sort(function(a, b) {
        const keyA = a[0];
        const keyB = b[0];

        if (keyA == keyB) return 0;


        if (reversed) {
            if (numOrFloat(keyA) && numOrFloat(keyB)) {
                return (keyB-keyA)>0 ? 1 : -1;
            }
            return keyA < keyB ? 1 : -1;
        } else {
[...]

PR #464, opened 2021-08-13
Add a 'keyFloat' function to sort "Duration" column numerically

gscelsi:
... instead of lexicographically. This simple change solves #449 and #413.

**Note** This merge would fix sorting of the *default* contents of the "duration" column. However, this merge would make sorting **not** work if users change formatting following instructions in the User's Guide [Formatting the Duration Column](https://pytest-html.readthedocs.io/en/latest/user_guide.html#formatting-the-duration-column) section. In that case the current lexicographic sorting would be required. Therefore at this point I think the best solution would be to decline this pull request and to change the default formatting of the "duration" column to "%H:%M:%S.%f".


BeyondEvil:
Duration is handled correctly now by v4.

PR #471, opened 2021-11-02
Update user_guide.rst

acohen716:
Since v3.0.0 HTML generation is post processed (https://pytest-html.readthedocs.io/en/latest/changelog.html#id13) and thus the values in the "Time" column don't represent the test end time, this fixes that


BeyondEvil:
I believe this is solved in v4.

PR #472, opened 2021-12-02
Update html_report.py

Raorajeevyadav:
This line will add passing percentage of the test cases in the summery. By this,Business Team have more clarity of the passing percentage.


BeyondEvil:
Thank you for this!

We're currently in the process of re-writing this plugin, so this feature will be considered for that.

BeyondEvil:
This should probably be added by the user using the `pytest_html_results_summary` hook.

PR #473, opened 2021-12-02
Update style.scss

Raorajeevyadav:
More dark color are used for content font and reporting table. Earlier fade color was used. Report title has been pushed to center to show more interactive. h3 and h4 header introduced  to show the passing percentage in green color.


Raorajeevyadav:
updating again

PR #484, opened 2022-01-17
Add reusable workflow

BeyondEvil:
GitHub Actions recently introduced [reusable workflows](https://github.blog/2021-11-29-github-actions-reusable-workflows-is-generally-available/).

I want to use that functionality to split out the different workflows to make them a little bit more focused and easier to understand.

This is a prepatory step to enable that.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,132 @@
+name: Tests
+
+on:
+  workflow_call:
+
+jobs:
+  test_python:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            name: py36-ubuntu
+            python-version: 3.6
+
+          - os: windows-latest
+            name: py36-windows
+            python-version: 3.6
+
+          - os: macOS-latest
+            name: py36-mac
+            python-version: 3.6
+
+          - os: ubuntu-latest
+            name: py37-ubuntu
+            python-version: 3.7
+
+          - os: windows-latest
+            name: py37-windows
+            python-version: 3.7
+
+          - os: macOS-latest
+            name: py37-mac
+            python-version: 3.7
+
+          - os: ubuntu-latest
+            name: py38-ubuntu
+            python-version: 3.8
+
+          - os: windows-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,132 @@
...
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Install Dependencies
+        run: npm install
+      - name: QUnit Tests
+        run: npm test
+
+  linting:
```
RonnyPfannschmidt, line 120: coudl thisone be migrated to precommit.ci?

RonnyPfannschmidt (approved):

.github/workflows/tests.yml
nicoddemus: https://pre-commit.ci/

We have been enabling this for many pytest plugins and are very happy with the results.

BeyondEvil: > https://pre-commit.ci/
> 
> We have been enabling this for many pytest plugins, including pytest itself, and are very happy with the results.

Oh, nice!

I'll follow up this PR with one to migrate it then. 👍

BeyondEvil: > https://pre-commit.ci/
> 
> We have been enabling this for many pytest plugins, including pytest itself, and are very happy with the results.

How exactly do I migrate?

According to pre-commit.ci docs - nothing is needed except the config file (which exists) and I couldn't find anything in the pytest repo to hint at how to do it either? @nicoddemus

BeyondEvil: > https://pre-commit.ci/
> 
> We have been enabling this for many pytest plugins, including pytest itself, and are very happy with the results.

Ah, I think I figured it out.

nicoddemus: You don't need to add anything to the repository, we just need to enable it in the project settings. Is that what you did?

nicoddemus: Yes:

image

BeyondEvil: > You don't need to add anything to the repository, we just need to enable it in the project settings. Is that what you did?

Yep, that's what I did. :)

PR #487, opened 2022-01-17
Migrate to precommit.ci

BeyondEvil:


nicoddemus (approved):
I see there's still a `linting` job running in there though.

BeyondEvil:
> I see there's still a `linting` job running in there though.

Yes, that's because the action uses an [absolute path](https://github.com/pytest-dev/pytest-html/blob/master/.github/workflows/actions.yml#L34) to the `tests.yml` and not the one in the PR.

As far as I can tell, unfortunately, you can't use a relative path.

BeyondEvil:
I see. All valid points. But at this time I feel that if it’s good enough for the main pytest project it’s good enough for us.

With limited time, at least I have more pressing issues to deal with.

Speaking of, any thoughts on why the windows tests started failing?

BeyondEvil:
FYI, you can now use relative paths. 😅

ssbarnea:
Sorry but Windows is the only platform about which I do not care, so I did not look at it. I already wasted almost two weeks making Ansible work under GHA WSL1, ended up with an instant heachache when I see Windows ;)

BeyondEvil:
Yeah, I hear ya.

I have a strong urge to remove official support for windows.

If it works it works, and I can accept patches. But that’s about it.

I don’t own a PC and I have a M1 with no feasible means of running a VM yet.

ssbarnea:
Not that I expect you do work on it, but I have an M1 too and I installed Windows 11 arm64 edition under Parallels and works quite nice and FAST!. Just do not expect to be able to use podman/docker/wsl inside that VM because M1 does not allow nested-virtualization. In fact I even share my code repository across those systems, so I do not need to clone/sync code when testing under a different platform. I was planning to write a blog post about that, but time....

BeyondEvil:
Yeah, but parallels isn't free. And I don't have a need for it for work and I don't feel like paying for something w.r.t. unpaid OS work.

I'll just wait until the folks at VirtualBox catches up.

Still nice, will read that blog-post.

ssbarnea:
That will not catch-up that time, they announced that there are no plans for arm64 on virtualbox. Also [UTM](https://mac.getutm.app/) seems might be something you want to investigate as open-source (its a gui on top of qemu)

BeyondEvil:
Yeah, I tried UTM but couldn't get it to work. I might revisit.

Yeah, I know what they announced. But unless they want to become irrelevant they'll have to as Apple will move "everything" to ARM.

PR #488, opened 2022-01-17
Separate Nightly workflow

BeyondEvil:


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,9 @@
+name: Nightly tests
```
nicoddemus, line 1: Out of curiosity, why separate the workflows?

nicoddemus (approved):

.github/workflows/nightly.yml
BeyondEvil: Basically for the same reasons we don't put all python code in one file, logical separation, readability, maintainability, etc.

Which, imo, is especially true for CI config which has a tendency to rarely see changes once defined. So having it logically separated (I will separate out `publish` to it's own file as well) will help when someone has to make changes.

BeyondEvil: For a concrete example: https://github.com/pytest-dev/pytest-html/blob/master/.github/workflows/actions.yml#L36

All these `if` makes at least my head hurt. Can I wrap my head around it? Sure, but it will take some time. And 6 months later when I need to revisit, I'm going to need to go through that pain again.

PR #496, opened 2022-02-04
troubleshoot failing windows test

BeyondEvil:


BeyondEvil:
fixed in https://github.com/pytest-dev/pytest-html/pull/522

PR #498, opened 2022-02-26
Use pypy-3.8 for tests

BeyondEvil:


BeyondEvil:
fixed in https://github.com/pytest-dev/pytest-html/pull/522

PR #502, opened 2022-03-15
Multi byte character method name is garbled

### LINKED ISSUE #501: Multi byte character method name is garbled
## issue

When use unicode in test method,

```python
    def test_あいうえお(self):
        print('かきくけこ')
```

I got garbled text (mojibake) in html reports.

```html
test_ããããã
```

## reason

Method name ( report.nodeid ) is encoded as utf-8 and decoded as unicode_escape .

**src/pytest_html/result.py**

```python
        self.test_id = report.nodeid.encode("utf-8").decode("unicode_escape")
```

Multi byte character as utf-8 is different same one as unicode_escape.

```python
>>> print('abcdeあいうえお'.encode("utf-8"))
b'abcde\xe3\x81\x82\xe3\x81\x84\xe3\x81\x86\xe3\x81\x88\xe3\x81\x8a'
>>> print('abcdeあいうえお'.encode("unicode_escape"))
b'abcde\\u3042\\u3044\\u3046\\u3048\\u304a'
```

kannkyo:
close #501

### DIFF
--- a/src/pytest_html/result.py
+++ b/src/pytest_html/result.py
@@ -18,7 +18,7 @@ from .util import ansi_support
 
 class TestResult:
     def __init__(self, outcome, report, logfile, config):
-        self.test_id = report.nodeid.encode("utf-8").decode("unicode_escape")
+        self.test_id = report.nodeid
         if getattr(report, "when", "call") != "call":
             self.test_id = "::".join([report.nodeid, report.when])
         self.time = getattr(report, "duration", 0.0)


BeyondEvil:
I wonder why the transformation from utf-8 to escape was added in the first time.

Let's see what happens with the tests. (pypy3 is failing but that's not your fault.).

BeyondEvil:
Can you rebase master @kannkyo ? 🙏

kannkyo:
@BeyondEvil I rebased branch.

BeyondEvil:
> @BeyondEvil how about ? I'll wait for fixed new pytest-html.

Yes, thank you for your patience!

BeyondEvil:
Should be fixed in v4.

PR #512, opened 2022-04-21
Original test cases order kept

sygutss:
Because that sometimes somebody
could want to have a possibility to
keep the original order of test cases
in the report, that possibility was
added. The logic is control by specific
flag in config file.

1. Added necessary option flag in config.
2. Added necessary logic if flag is set to True.
3. Added necessary tests for new logic.
4. Added additional section in documentation
which describes how to turn on the new behaviour.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -261,6 +261,21 @@ or by setting the :code:`render_collapsed` in a configuration file (pytest.ini, 
 
 **NOTE:** Setting :code:`render_collapsed` will, unlike the query parameter, affect all statuses.
 
+
+Keep original test cases run order
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, all rows in the **Results** table will be sorted by category. Failed test cases will be placed at the top of the **Results** table.
+
+This behavior can be customized by setting the :code:`keep_original_order` in a configuration file (pytest.ini, setup.cfg, etc).
+
+.. code-block:: ini
+
+  [pytest]
+  keep_original_order = True
+
+**NOTE:** Setting :code:`keep_original_order` will turn off a possibility of changing order by table headers.
+
 Controlling Test Result Visibility Via Query Params
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/pytest_html/html_report.py
+++ b/src/pytest_html/html_report.py
@@ -36,15 +36,20 @@ class HTMLReport:
     def _appendrow(self, outcome, report):
         result = TestResult(outcome, report, self.logfile, self.config)
         if result.row_table is not None:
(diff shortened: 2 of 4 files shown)


sygutss:
[Question]
Because of the fact that right now there is a problem with the current black version (Linters are failed) Should I also provide fix for that under that PR? The latest version of black works fine and all pre-commit checks are passed.

BeyondEvil:
Thanks for the PR. 🙏 

I'll fix black. 👍

BeyondEvil:
Try rebasing with master now.

sygutss:
Rebased and black formatting fixes made by black

sygutss:
See that tests for py37-windows failed, dunno why... Can we re-run that checks? I closed and re-opened the pull request, probably checks start is waiting for some approval.

BeyondEvil:
And now it failed for py38-windows.

I might just add re-run flaky for windows.

zwibldog:
Hi guys,  

this feature looks really great and is exactly what I'm looking for. @sygutss @BeyondEvil any update in this PR? See any chance to merge it in near future? Thanks!

BeyondEvil:
As soon as 4.0.0 is GA (4.0.0rc0 out now!) I'll revisit this.

BeyondEvil:
@sygutss @zwibldog This should be the default behaviour in `4.0.0rc1`.

Please test and report any issues. 🙏

sygutss:
Hi @BeyondEvil I've updated to 4.0.0rc1 but it's not like expected in default behavior. Behavior is like the previous approach - By default we have firstly error cases and after them passed cases. After clicking on ordering in that column (header) we can switch opposite - First passed cases after failed cases. There is a lack of the possibility to achieve the original ordering

Conclusion - the original ordering is not the default behavior.

BeyondEvil:
> 

Ok, thanks. I think I initially misunderstood the ask.

But OK, I get what you want now.

This happens to coincide with https://github.com/pytest-dev/pytest-html/issues/521

So I'll see if I can add a query param to set initial sort order (column, asc/desc, and "original" or something).

BeyondEvil:
Check out #637 and #634 

They will be part of the 4.0.0rc2 release coming soon.

BeyondEvil:
Please try 4.0.0rc4. setting query param ”sort” to ”original” will now sort in execution order.

sygutss:
Hi - I've tested it - in version 4.0.0rc4 and setting query param ”sort” to ”original” the ordering is as expected :)

Additional question - will be possible to enforce that behavior as a default one? eg. from the config file? I mean to always open by default the result html file with that ordering without a necessity to setting different ordering by query param manually?

sygutss:
@BeyondEvil - Up for the previous question and also when is expected to roll out the official version?

BeyondEvil:
Sorry for the delay. This can be set using addopts and/or ini. And v4 is out.

Thanks! 🙏

PR #514, opened 2022-05-05
Use pathlib.Path for file path handling

mrlegohead0x45:
This PR changes all file and directory path handling to use `pathlib.Path` instead of using strings and `os.path` functions. The initial reason for this is that simply prepending `"file://"` to the path of the report does not generate a valid URI on Windows, where the directory separator is `\`, because you end up with something like `file://D:\proj\report.html`, which uses backslashes, and file URIs are [supposed to use forward slashes](https://en.wikipedia.org/wiki/File_URI_scheme#:~:text=Forward%20slashes%20should%20be%20used%20to%20delimit%20paths.). Instead we can use the [`pathlib.Path.as_uri()`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_uri) function to get a valid URI. While I was at it, I changed the rest of the file[s] to use `pathlib.Path` as well.

### DIFF
--- a/src/pytest_html/html_report.py
+++ b/src/pytest_html/html_report.py
@@ -87,9 +87,7 @@ class HTMLReport:
         numtests = self.passed + self.failed + self.xpassed + self.xfailed
         generated = datetime.datetime.now()
 
-        with open(
-            os.path.join(os.path.dirname(__file__), "resources", "style.css")
-        ) as style_css_fp:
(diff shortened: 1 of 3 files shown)


BeyondEvil:
Thank you for this!

We've included this in the next-gen project (v4).

We'll make sure to credit you for the contribution. 🙏

BeyondEvil (approved):
LGTM!

PR #522, opened 2022-07-03
fix: upgrade pip before running tox

BeyondEvil:


.github/workflows/tests.yml
```diff
@@ -63,9 +63,10 @@ jobs:
             name: pypy3-ubuntu
             python-version: pypy-3.8
 
-          - os: windows-latest
-            name: pypy3-windows
-            python-version: pypy-3.8
+          # TODO: This test takes 10(!) times as long as the regular py38 on Windows
+#          - os: windows-latest
+#            name: pypy3-windows
+#            python-version: pypy-3.8
```
mattip, line 69: wow, that is slow.I wonder if pypy-3.8-nightly is any faster?

BeyondEvil: I can try

BeyondEvil: Still slow @mattip https://github.com/pytest-dev/pytest-html/runs/7377268119?check_suite_focus=true

mattip: wow, 16m 26s for pypy3.8 on windows, ~2m for cpython on windows, ~3 min for pypy on ubuntu, ~1 min for cpython on ubuntu.

BeyondEvil: I wonder what exactly is happening.

A lot of tests use regex, could that be it?

Anyway, I'm rewriting the plugin, and as an effect of that, I'm also having to re-write a lot of the test.

It will be interesting to see what the difference might be.

PR #546, opened 2022-09-01
Enabled numeric sorting of the 'duration' column

rindPHI:
This PR makes the "duration" column *numerically* sortable, such that, e.g., "2.0" appears *before* "100.0" and not after as in an alphabetic sort. I consider this very important to track down performance regressions as witnessed by test case execution times.

### DIFF
--- a/src/pytest_html/html_report.py
+++ b/src/pytest_html/html_report.py
@@ -146,7 +146,7 @@ class HTMLReport:
         cells = [
             html.th("Result", class_="sortable result initial-sort", col="result"),
             html.th("Test", class_="sortable", col="name"),
-            html.th("Duration", class_="sortable", col="duration"),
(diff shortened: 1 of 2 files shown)


BeyondEvil:
Thank you for the PR and contribution.

We're currently working hard on delivering the "next-gen" version of this plugin.

I think this is already supported by the new version.

As soon as the beta is released, I'll ping here and I would much appreciated it if you could test.

rindPHI:
That sounds great, looking forward to testing the new version!

mnaumann-plenty:
This is long overdue -- I am SO looking forward to this...

BeyondEvil:
@rindPHI @mnaumann-plenty Mind trying out 4.0.0rc1 ?

BeyondEvil:
This should be supported in v4 now.

PR #550, opened 2022-09-30
fix: multi byte character method name

### LINKED ISSUE #501: Multi byte character method name is garbled

wf-yamaday:
## What this?

This PR resolves  #501 .

## Related PR

#502 is the same solution. But the unit test `test_report_display_utf8` is fails.
So, I also change the unit test to have the expected behavior.

### DIFF
--- a/src/pytest_html/result.py
+++ b/src/pytest_html/result.py
@@ -18,7 +18,7 @@ from .util import ansi_support
 
 class TestResult:
     def __init__(self, outcome, report, logfile, config):
-        self.test_id = report.nodeid.encode("utf-8").decode("unicode_escape")
+        self.test_id = report.nodeid
         if getattr(report, "when", "call") != "call":
             self.test_id = "::".join([report.nodeid, report.when])
         self.time = getattr(report, "duration", 0.0)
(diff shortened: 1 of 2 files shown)


BeyondEvil:
Updated in #760

PR #554, opened 2022-10-25
Add py dependency as pytest removed it in 7.2.0

PeterStolz:


PeterStolz:
Ah confused build system with

PR #561, opened 2022-10-27
Update pre-commit hooks

ssbarnea:


BeyondEvil (approved):
Half of this code will be removed/changed once we go v4. But I guess it doesn't matter for now.

BeyondEvil:
Obviously make sure tests pass before we merge.

BeyondEvil:
> @BeyondEvil Any idea about when you expect v4? I started this only because i spotted some "degradation" on CI... Maybe I can avoid doubling work.

We're _basically_ at 1:1 with the current state. The only thing that I can think of that we're missing is ANSI-support. And I hear you're the expert ;)

This is the branch with the most current work: https://github.com/BeyondEvil/pytest-html/tree/combined-fe-and-be

I plan on merging that with the `next-gen` branch before EOW. Once merged, we release v4.0.0-RC0. @ssbarnea

BeyondEvil:
Think this is handled by the pre-commit auto update.

PR #564, opened 2022-11-04
Fix: The report is saved without following the UTF-8 encoding

dongfangtianyu:
HTML specifies the page charset="utf-8", but the file was not actually save according to utf-8

## How to reproduce 
a case
```python

def test_abc():
    print("这是UTF-8内容") 

```

## Report content
image

## Cause analysis

This line of code declares that the HTML content is displayed in UTF-8, but in fact, the HTML file is not saved in UTF-8.

```html
<meta charset="utf-8"/>
```

To put it simply, the encoding of the file content is unknown, but it is emphasized to display according to UTF-8

#224

### DIFF
--- a/src/pytest_html/html_report.py
+++ b/src/pytest_html/html_report.py
@@ -256,10 +256,10 @@ class HTMLReport:
         if not self.self_contained:
             assets_dir.mkdir(parents=True, exist_ok=True)
 
-        self.logfile.write_text(report_content)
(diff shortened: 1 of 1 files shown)


BeyondEvil:
Mind testing if 4.0.0rc1 solves your issue?

BeyondEvil:
v4 seems to handle this correctly.

PR #567, opened 2022-11-11
Switch to setuptools-scm >= 7.0.0

dvzrv:
pyproject.toml:
Switch to setuptools-scm >= 7.0.0, which allows dropping setuptools-scm-git-archive.
Remove wheel, as it is not required for building using PEP517 tooling (i.e. setuptools, setuptools-scm, build, installer).

setup.py:
Raise required setuptools-scm version to match the build-system requirements in pyproject.toml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,7 @@
 [build-system]
 requires = [
-  "setuptools >= 42",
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):
Lovely, thanks

ssbarnea (approved):

Issue #566, opened 2022-11-11
when time < 10ms, the duration will be 0


WangFengtu1996:
* The duration_formatter is set `%M:%S.%f`
```
@pytest.hookimpl(hookwrapper=True)
def pytest_runtest_makereport(item, call):
    '''data from the output of pytest gets processed here
     and are passed to pytest_html_results_table_row'''
    outcome = yield
    # this is the output that is seen end of test case
    report = outcome.get_result()
    setattr(report, "duration_formatter", "%M:%S.%f")
```
* got the duration will be zero.
![image](https://user-images.githubusercontent.com/21277368/201329525-e829e94b-e1af-427c-bb81-be7fd81602cb.png)

* This may not be bug. would someone give me some guide to show more detail time?
* Thanks for your time.

BeyondEvil:
I think it depends on what we get from pytest.

I'll have a look, but I don't think you can get a better resolution on v3. But I'll see what the possibilities are for v4 (next-gen).

BeyondEvil:
Having spent almost 2 hours looking at both the old and new duration formatting - I'm **VERY** tempted to remove this functionality all together.

Replace it with something that is not configurable, but "smarter":

"10 ms"

"34 sec"

"6:13 min"

"1:45:26 hrs"

Do you have a use-case where the formatting matters, apart from not showing 0 for a <10ms test?

TUT-jiayou:
Hi，I&nbsp;just&nbsp;test&nbsp;the&nbsp;same&nbsp;case&nbsp;by&nbsp;adding&nbsp;sleep&nbsp;10ms&nbsp;then&nbsp;I&nbsp;get&nbsp;number&nbsp;00:00.01.&nbsp;I&nbsp;add&nbsp;sleep&nbsp;1s,&nbsp;then&nbsp;get&nbsp;00:01.00.&nbsp;

---Original---
From: &quot;Jim ***@***.***&gt;
Date: 2022/11/11 22:10:21
To: ***@***.***&gt;;
Cc: &quot;Fengtu ***@***.******@***.***&gt;;
Subject: Re: [pytest-dev/pytest-html] when time < 10ms, the duration will be 0(Issue #566)




 
Having spent almost 2 hours looking at both the old and new duration formatting - I'm VERY tempted to remove this functionality all together.
 
Replace it with something that is not configurable, but "smarter":
 
"10 ms"
 
"34 sec"
 
"6:13 min"
 
"1:45:26 hrs"
 
[...]

PR #568, opened 2022-11-16
modify feature when time < 10ms, the duration will be 0

WangFengtu1996:
more detail https://github.com/pytest-dev/pytest-html/issues/566


BeyondEvil:
This functionality has been deprecated for the next major release.

The timestamps are going to be `nnn ms` for everything <1 second and `HH:mm:ss` for everything  >=1 second.

If a need arises, we'll provide a hook where you can format it yourself.

PR #573, opened 2022-11-23
Large img preview in self-contained mode

mganisin:
"ful-screen" preview of extra images attached to results for self-contained mode. Unlike image preview in normal mode implemented as part of page with little bit of js+css (likely also possible to implement via <a>, however adress line populated by base64 of the image would be really ugly in such case).

### DIFF
--- a/src/layout/css/style.scss
+++ b/src/layout/css/style.scss
@@ -141,6 +141,7 @@ div.image {
 	margin-left: $spacing;
 	overflow: hidden;
 	width: $extra-media-width;
+	cursor: zoom-in;
 
 	img {
 		width: $extra-media-width;
@@ -213,5 +214,30 @@ div.video {
 	.desc.active & {
 		/*finish triangle*/
 		border-top: $triangle-width solid #999;
+	}
+}
+
+
+/*---------------------------
+ * 4. Zoom in self-contained
+ *---------------------------*/
+
+#large {
+	display: none;
+	top: 0;
+	left: 0;
+	width: 100%;
+	height: 100%;
+	z-index: 100;
+	position: fixed;
+	margin: auto;
+	overflow: scroll;
+	cursor: zoom-out;
+	background: rgba(0,0,0,0.5) no-repeat center;
+
+	img {
+		display: block;
+		margin-left: auto;
+		margin-right: auto;
 	}
 }
--- a/src/pytest_html/html_report.py
+++ b/src/pytest_html/html_report.py
@@ -201,6 +201,9 @@ class HTMLReport:
 
         body.extend(results)
 
+        if self.self_contained:
+            body.append(html.div(html.img(), id="large", onclick="zoomOut()"))
+
         doc = html.html(head, body)
 
         unicode_doc = "<!DOCTYPE html>\n{}".format(doc.unicode(indent=2))
--- a/src/pytest_html/resources/main.js
+++ b/src/pytest_html/resources/main.js
@@ -111,6 +111,23 @@ function addCollapse() {
     });
 }
 
+function zoomIn(event) {
+    const large = find('#large');
+    find('img', large).setAttribute('src', event.target.getAttribute('src'));
+    large.style.display = 'block';
+    large.scrollTop = 0;
+}
+
+function zoomOut() { // eslint-disable-line no-unused-vars
+    find('#large').style.display = 'none';
+}
+
+function addZoomIn() {
+    findAll('.image img').forEach(function(elem) {
+        elem.addEventListener('click', zoomIn);
+    });
+}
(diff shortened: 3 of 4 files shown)


BeyondEvil:
Thanks for the PR!

Since we're about to release the first RC of v4 (aka next-gen) which is a complete rewrite of the plugin, let's put this on hold.

One of the things we've implemented in v4 is an image gallery, and I think this has been solved there.

BeyondEvil:
Can you try `4.0.0rc0` and see if it works as expected?

mganisin:
> Can you try `4.0.0rc0` and see if it works as expected?

Could be a breaking change there? I get:

```
INTERNALERROR>     extra.append(pytest_html.extras.url(issue, name=issue_id))                                                           INTERNALERROR> AttributeError: 'function' object has no attribute 'url'
```

mganisin:
> > > Can you try `4.0.0rc0` and see if it works as expected?
> > 
> > 
> > Could be a breaking change there? I get:
> > ```
> > INTERNALERROR>     extra.append(pytest_html.extras.url(issue, name=issue_id))
> > INTERNALERROR> AttributeError: 'function' object has no attribute 'url'
> > ```
> 
> Can you try ”from pytest_html import extras” instead?

Right, I found the problem (I used pluginmanager to get pytest-html). This doesn't seem to be only breaking change, At least `report.extra` is renamed to `report.extras` without backward compatibility (my preference is to not break interfaces, but I understand that not everyone shares this approach).

screenshots in self-contained mode don't seem to work at all (same code used with old and new version, got these results):
1. `src` attribute of `` is different. current version adds `data:image/png;base64,${BASE64}` version 4.0.0 adds just `${BASE64}`. Preview screenshot is not displayed in new version (while properly displayed in old)
2. Clicking the preview (which is not displayed) opens URL with `${BASE64}` included (I tried this as well when working on images preview as this is easiest way to do that, but found base64 in URL pretty inconvenient; if I remember properly, my attempt worked, maybe because of `data:...` prefix which is missing in 4.0.0). That results in error:
 
```
Request-URI Too Long

The requested URL's length exceeds the capacity limit for this server.
```

Footnote: Personally I don't appreciate new data-jsonblob. pytest-html is amazing also because of the speed. This won't improve the speed (not to mention that such result is unprocessable).

BeyondEvil:
This is awesome feedback, thank you! 🙏 

> > 
> > Can you try ”from pytest_html import extras” instead?
> 
> Right, I found the problem (I used pluginmanager to get pytest-html). This doesn't seem to be only breaking change, At least `report.extra` is renamed to `report.extras` without backward compatibility (my preference is to not break interfaces, but I understand that not everyone shares this approach).

Correct. I'm exploring ways of keeping the "pluginmanager" functionality. But I'd like to at least warn when it's used. That type of usage makes future changes difficult. The proper way is to use what I recommended, a regular import.

As for "report.extra", that should be backwards compatible, but throws a `DeprecationWarning` when used. Could you please share how it's breaking for you? 

> screenshots in self-contained mode don't seem to work at all (same code used with old and new version, got these results):
> 
> 1. `src` attribute of `` is different. current version adds `data:image/png;base64,${BASE64}` version 4.0.0 adds just `${BASE64}`. Preview screenshot is not displayed in new version (while properly displayed in old)
> 2. Clicking the preview (which is not displayed) opens URL with `${BASE64}` included (I tried this as well when working on images preview as this is easiest way to do that, but found base64 in URL pretty inconvenient; if I remember properly, my attempt worked, maybe because of `data:...` prefix which is missing in 4.0.0). That results in error:
> 
> ```
> Request-URI Too Long
> 
> The requested URL's length exceeds the capacity limit for this server.
> ```
> 

I'm having trouble reproducing at least 1. In my testing I get previews in both regular and self-contained mode. Would you mind sharing some reproducible example code/test?

> Footnote: Personally I don't appreciate new data-jsonblob. pytest-html is amazing also because of the speed. This won't improve the speed (not to mention that such result is unprocessable).

Would you mind elaborating on this a little bit?

How many tests does your report contain? Do all of them have extras in the form of images? In what way is the report slow? Rendering? Scrolling? etc.

Also, as for the unprocessable bit, what are you doing today that isn't possible with the new report?

My intent with v4 is that it should strive to be 1-to-1 with the old report as mush as possible. I'm trying my best to make as much as possible backwards compat. For things I want to change/remove I'm doing DeprecationWarnings so that I in v5 can remove them.

If you check the docs, all deprecations should be listed and have suggested mitigations.

Again, thank you for input - it's super valuable! 🙏

BeyondEvil:
Proper v4 is released now if you're interested in revisiting this @mganisin ?

PR #591, opened 2023-03-10
fix: Report fails to render with pytest-xdist (1)

BeyondEvil:
The reason why the report fails to render when running with `pytest-xdist`, is that it adds a `node` attribute with a `WorkerController` object as its value.

This object is unserializable, so it fails when trying to convert the test data to JSON.

This is proposed solution 1.

Solution 2: #592 
Solution 3: #593 

There's also the option of making the WorkerController class serializable upstream. Ping @nicoddemus

### DIFF
--- a/src/pytest_html/nextgen.py
+++ b/src/pytest_html/nextgen.py
@@ -239,14 +239,19 @@ class BaseReport:
 
     @pytest.hookimpl(trylast=True)
     def pytest_runtest_logreport(self, report):
-        data = self._config.hook.pytest_report_to_serializable(
+        serialized_report = self._config.hook.pytest_report_to_serializable(
             config=self._config, report=report
         )
+
+        data = {
+            "duration": serialized_report["duration"],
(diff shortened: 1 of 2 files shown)


nicoddemus:
> There's also the option of making the WorkerController class serializable upstream. Ping @nicoddemus

IMHO the WorkerController is an implementation detail and of little to no value to users, so I don't see an advantage in serializing it. :+1:

nicoddemus:
Possibly, but more investigation would be needed.

BeyondEvil:
> Possibly, but more investigation would be needed.

I think it's safe to keep it out of the report.

It wasn't part of the legacy report afaict. And if someone has a use case for having it, we can re-add it easily.

RonnyPfannschmidt (changes requested):
this is lossy in the face of valid extensions, os i recommend against it

BeyondEvil:

PR #597, opened 2023-03-18
refactor: stop overwriting pytest data

BeyondEvil:
Previously we were overwriting some of the attributes coming out of pytest (like `nodeid`) and since this data is shared with other plugins, that's a bad idea.

This PR aims to fix that.


RonnyPfannschmidt (commented):
is there any reason for report to use dictly typing?

as for updates to the serialzed report details, it may make sense to add a enw attribute to the actual report and only serializing iz atthe end

PR #603, opened 2023-03-25
fix: Missing logging in report

BeyondEvil:
The logging output/capture was broken in the legacy report, where "teardown" logs wasn't visible in the tests log output.

### DIFF
--- a/src/pytest_html/nextgen.py
+++ b/src/pytest_html/nextgen.py
@@ -66,10 +66,18 @@ class BaseReport:
                 "collectedItems": 0,
                 "runningState": "not_started",
                 "environment": {},
-                "tests": [],
+                "tests": defaultdict(list),
                 "resultsTableHeader": {},
                 "additionalSummary": defaultdict(list),
             }
+
+        @property
+        def title(self):
+            return self._data["title"]
+
+        @title.setter
+        def title(self, title):
+            self._data["title"] = title
 
         @property
         def config(self):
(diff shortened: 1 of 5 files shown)


src/pytest_html/nextgen.py
```diff
@@ -66,11 +66,19 @@ def __init__(self, title, config):
                 "collectedItems": 0,
                 "runningState": "not_started",
                 "environment": {},
-                "tests": [],
+                "tests": defaultdict(list),
```
BeyondEvil, line 69: Made a small performance improvement to avoid having to iterate through all tests, we key them with the original `nodeid`.

src/pytest_html/nextgen.py
```diff
@@ -291,14 +313,11 @@ def pytest_runtest_logreport(self, report):
...
-        # Order here matters!
-        log = report.longreprtext or report.capstdout or "No log output captured."
-        data["log"] = _handle_ansi(log)
         data["result"] = _process_outcome(report)
         data["extras"] = self._process_extras(report, test_id)
 
-        self._report.add_test(data)
-        self._generate_report()
+        if self._report.add_test(data, report):
+            self._generate_report()
```
BeyondEvil, line 320: Only generate the report if we actually made a change (added a test in this case).

src/pytest_html/scripts/datamanager.js
```diff
@@ -3,7 +3,7 @@ const { getCollapsedCategory } = require('./storage.js')
 class DataManager {
     setManager(data) {
         const collapsedCategories = [...getCollapsedCategory(), 'passed']
-        const dataBlob = { ...data, tests: data.tests.map((test, index) => ({
+        const dataBlob = { ...data, tests: Object.values(data.tests).flat().map((test, index) => ({
```
BeyondEvil, line 6: This flattens (removes the nodeid key) from the JSON data.

src/pytest_html/scripts/dom.js
```diff
@@ -84,6 +84,7 @@ const dom = {
         formattedDuration = formatDuration < 1 ? formattedDuration.ms : formattedDuration.formatted
         const resultBody = templateResult.content.cloneNode(true)
         resultBody.querySelector('tbody').classList.add(resultLower)
+        resultBody.querySelector('tbody').id = testId
```
BeyondEvil, line 87: Added this to be able to differentiate between logs using the `testId`.

PR #625, opened 2023-04-05
chore: It's <current year>, 120 is fine

BeyondEvil:


RonnyPfannschmidt (commented):
black is very clean on why 88 ^^

BeyondEvil:
I'll go with the Bugbear version, I don't want to be bothered if my line of text is 89, because then I have to introduce awkward breaks in the text.

For code, 88 is fine.

PR #629, opened 2023-04-07
Fix: Order and layout of outcome summary

BeyondEvil:
Making the order of the outcomes in the summary the same as `pytest` has in the terminal.


src/pytest_html/scripts/main.js
```diff
@@ -62,10 +62,15 @@ const renderContent = (tests) => {
 
 const renderDerived = (tests, collectedItems, isFinished) => {
     const currentFilter = getVisible()
-    possibleResults.forEach(({ result, label }) => {
+    possibleResults.forEach(({ result, label }, index) => {
         const count = tests.filter((test) => test.result.toLowerCase() === result).length
```
drRedflint, line 66: Why include index?

BeyondEvil: Good catch, it's a left-over from a previous solution. Will remove.

PR #703, opened 2023-07-28
Fix: Collapsed state between redraws

BeyondEvil:


src/pytest_html/scripts/datamanager.js
```diff
@@ -1,19 +1,25 @@
...
 
 class DataManager {
     setManager(data) {
         const collapsedCategories = [...getCollapsedCategory(data.renderCollapsed)]
+        const collapsedIds = []
         const tests = Object.values(data.tests).flat().map((test, index) => {
             const collapsed = collapsedCategories.includes(test.result.toLowerCase())
+            const id = `test_${index}`
+            if (collapsed) {
+                collapsedIds.push(id)
```
drRedflint, line 11: I'd avoid updating the collapsedIds collection in the map.
Instead I would do it after the construction of collapsed collection:
`const collapsedIds = collapsed.map(({id}) => id)`

drRedflint: or even better: 
setCollapsedIds(collapsed.map(({id}) => id))

BeyondEvil: I actually had it that way at first. But thought it was better to not do two iterations over the same collection.

src/pytest_html/scripts/datamanager.js
```diff
@@ -47,6 +53,10 @@ class DataManager {
     get environment() {
         return this.renderData.environment
     }
+
+    get initialSort() {
+        return this.data.initialSort
```
drRedflint, line 58: `return [...this.data.initialSort]`

This prevents data.initialSort from becoming compromised by being updated by reference.

PR #749, opened 2023-10-17
Fix results table modification documentation

michalkaptur:
The example results table modification shown in the docs is broken since https://github.com/pytest-dev/pytest-html/pull/599, some hardcoded strings are used instead of the dynamically generated report entry data.


BeyondEvil (approved):
LGTM!

PR #789, opened 2024-01-08
Fix license classifier in pyproject.toml

tapdiego-amzn:
Project is licensed as MPL-2.0 but a MIT license classifier is being used, resulting in the wrong license being displayed on PyPI. Changing classifier to the correct one for the actual project license.


BeyondEvil:
Nice, thank you for this!

Issue #806, opened 2024-04-09
.col-results.collapsed issue if results table order changed


volkan-aslan:
### Discussed in https://github.com/pytest-dev/pytest-html/discussions/801

<div type='discussions-op-text'>

<sup>Originally posted by **volkan-aslan** March  7, 2024</sup>
Hello,

My pytest-html version is: 4.1.1

I am using below code in my conftest file to move results section to right-hand side.

```
def pytest_html_results_table_header(cells):
    cells.append(cells.pop(0))

def pytest_html_results_table_row(report, cells):
    cells.append(cells.pop(0))
```

But i noticed that, doing this breaking the following part of the report and it always shows "hide details" even its collapsed.
(This was not a case with version 3.2.0),

```
.col-result:hover::after {
  content: " (hide details)";
}

.col-result.collapsed:hover::after {
  content: " (show details)";
}

```

When I inspect the html report, I noticed that it's adding the 'collapsed' class to the first td of .collapsible.
This is OK if col-result is the first row, but in my case i moved it to the last so first row is col-testId, therefore it adds collapsed to col-testId.
```
if (collapsed) {
            resultBody.querySelector('.collapsible > td')?.classList.add('collapsed')
            resultBody.querySelector('.extras-row').classList.add('hidden')
        } else {
            resultBody.querySelector('.collapsible > td')?.classList.remove('collapsed')
        }
```
If I change it to:
**resultBody.querySelector('.collapsible > col-result')?.classList.add('collapsed')**

then it works as expected.

I am not a java script person, so I don't know if this is the ideal fix but it would be nice to have a solution for such a case.
If there's anything I've done wrong, I'd appreciate it if you could help.

Thanks.

</div>

![ss](https://github.com/pytest-dev/pytest-html/assets/83580341/c13155dc-15b6-4330-b366-33eecf0a57ee)

volkan-aslan:
https://github.com/pytest-dev/pytest-html/pull/807

BeyondEvil:
Thank you for reporting this. It is a known issue, but was never high on the todo because I never in a million years thought anyone would want the result column anywhere but first. 😆

PR #807, opened 2024-04-09
Fix for #806 (if results table order changed, hide/show details hover over broken)

volkan-aslan:
If you move 'col-result' to right hand side in results-table, it always shows 'hide details' even details are already collapsed.

the reason is, col-result is not the first td element of the results-table anymore and therefore it adds collapsible class to first td element.

because of this reason, below function does not work:
```
.col-result:hover::after {
  content: " (hide details)";
}

.col-result.collapsed:hover::after {
  content: " (show details)";
}

```

changing '.collapsible > td' to '.collapsible > col-result' fix this problem.

### DIFF
--- a/src/pytest_html/scripts/dom.js
+++ b/src/pytest_html/scripts/dom.js
@@ -57,10 +57,10 @@ const dom = {
         }
 
         if (collapsed) {
-            resultBody.querySelector('.collapsible > td')?.classList.add('collapsed')
+            resultBody.querySelector('.collapsible > col-result')?.classList.add('collapsed')
             resultBody.querySelector('.extras-row').classList.add('hidden')
         } else {
-            resultBody.querySelector('.collapsible > td')?.classList.remove('collapsed')
+            resultBody.querySelector('.collapsible > col-result')?.classList.remove('collapsed')
         }
 
         const media = []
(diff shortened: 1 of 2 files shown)


src/pytest_html/scripts/dom.js
```diff
@@ -57,10 +57,10 @@ const dom = {
         }
 
         if (collapsed) {
-            resultBody.querySelector('.collapsible > td')?.classList.add('collapsed')
+            resultBody.querySelector('.collapsible > col-result')?.classList.add('collapsed')
```
BeyondEvil, line 60: I don't have the code infront of my right now as I'm on vacation, but are we sure this will uniquely identify the column?

I ask because IIRC other things can have at least the ".collapsible" class.

BeyondEvil (changes requested):
It be great if you could add a test for this as well, preferably an integration test.

BeyondEvil:
Thank you for your contribution!

Please see my comments.

src/pytest_html/scripts/dom.js
volkan-aslan: I believe it is but better if you can double check when you return from vacation.
Thanks.

BeyondEvil (approved):
LGTM!

PR #812, opened 2024-05-09
Add expand_logs ini option

cidlik:
Is some cases it is more convenient to expand all logs by default,
e.g., if you have several tests with long logs and you need to find
something.


cidlik:
@BeyondEvil hello! Can you review this change, please?

PR #821, opened 2024-07-05
Modified the report style. I hope it's more beautiful

bomask:
My English is not good, so I use machine software to translate. I hope you can understand.

I made a bold attempt by modifying the style a bit, hoping it would become better. Including adjusting line spacing, font size, and color. And added a table. I hope adjusting the report style will also be a part of our future focus。

I tried my best to only change the style section without changing the code logic。

current report 
old

new:
new

### DIFF
--- a/src/pytest_html/resources/index.jinja2
+++ b/src/pytest_html/resources/index.jinja2
@@ -1,120 +1,183 @@
 <!DOCTYPE html>
 <html>
-  <head>
+<head>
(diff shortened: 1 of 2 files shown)


yuqinyao:
Bravo! It would be nice is information about percentage is added to the pie-chart.

BeyondEvil:
What happened with the `Environment` table?

PR #836, opened 2024-10-11
fix: support for pytest-retries

### LINKED ISSUE #815: Crashes with KeyError: 'retried' when test is retried with pytest-retry
It looks like the html reporter does not handle tests with the outcome `retried`. 

It's quite easy to reproduce:
1. Create a failing test with the mark `@pytest.mark.flaky(retries=3, delay=1)`
2. Run the test and wait for it to fail so that it is retried
3. Pytest crashes with:
```
File "/__w/xi/xi/tests/venv/lib/python3.9/site-packages/pytest_html/report_data.py", line 94, in outcomes
  self._outcomes[outcome.lower()]["value"] += 1
KeyError: 'retried'
```

acolombier:
Fixes #815


BeyondEvil:
Thanks for the PR.

Please rebase with master.

PR #842, opened 2024-11-19
Drop python 3.8, add support for 3.12, 3.13

Pierre-Sassoulas:


.github/workflows/tests.yml
```diff
@@ -87,21 +87,12 @@ jobs:
...
 
-          - os: ubuntu-latest
-            python-version: 3.11-dev
-            tox-env: devel
-          - os: windows-latest
-            python-version: 3.11-dev
-            tox-env: devel
-          - os: macos-latest
-            python-version: 3.11-dev
-            tox-env: devel
```
BeyondEvil, line 104: Why were these removed?

.github/workflows/tests.yml
```diff
@@ -163,12 +154,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "pypy3.9"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "pypy3.9"]
         include:
           - python-version: "3.10"
             with-coverage: true
-          - python-version: 3.11-dev
-            tox-env: devel
```
BeyondEvil, line 171: Dito.

.github/workflows/tests.yml
```diff
@@ -233,10 +222,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "pypy3.9"]
-        include:
-          - python-version: 3.11-dev
-            tox-env: devel
```
BeyondEvil, line 239: Dito.

.github/workflows/tests.yml
Pierre-Sassoulas: It's included in the new matrice with 3.11 (or 3.14-dev when it's going to be available), right ?

BeyondEvil: 3.11 yes.

But `devel` is a special target for tox where we run the suite agains pytest main-branch: https://github.com/pytest-dev/pytest-html/blob/master/tox.ini#L32

BeyondEvil: So set it to `3.13`.

BeyondEvil:
Most errors are related to `pkg_resources` which I guess is entirely removed in 3.12.

However, the code:

```python
    except AttributeError:
        # Needed for python < 3.9
        return pkg_resources.resource_string(
            "pytest_html", os.path.join("assets", "style.css")
        ).decode("utf-8")
```

We shouldn't hit.

So something is causing an attribute error here: https://github.com/pytest-dev/pytest-html/blob/master/testing/test_unit.py#L22

Pierre-Sassoulas:
Thanks for the review @BeyondEvil ! I did not manage to install locally yet (absolute path error I don't understand). Planning to work on this in the next few days and hopefully be able to reproduce the CI fail locally if adding setuptools doesn't work.

BeyondEvil:
> Thanks for the review @BeyondEvil ! I did not manage to install locally yet (absolute path error I don't understand). Planning to work on this in the next few days and hopefully be able to reproduce the CI fail locally if adding setuptools doesn't work.

Ugh... windows builds are always sooo slow.

Anyway, I think it's fixed.

Re-add the `devel` stuff and I think we're good to merge. 👍 

Thanks for the PR! ❤️ 🙏

Pierre-Sassoulas:
Nice ! Will do. I plan to merge all the dependabot one next, there's a bunch of pending cve warning that stressed me out :D

Pierre-Sassoulas:
Thank you for fixing the pipeline, I had a hard time setting up the dev env locally to troubleshoot !

PR #846, opened 2024-11-20
[pre-commit] Add pyproject-fmt to normalize the pyproject.toml

Pierre-Sassoulas:
Will prevent most issues when doing parallel modifications of the pyproject.toml conf. (Every changes in pyproject.toml is automated).


BeyondEvil:
Are the changes to pyproject.toml made by the formatter?

PR #861, opened 2025-02-04
Create a dependabot.yml to be explicit about what need to be updated

Pierre-Sassoulas:


Pierre-Sassoulas:
It seems now that a github action is outdated and dependabot is not upgrading which is what I'm trying to fix with this PR. Previousely there was an issue with a dependencie at build time that I tried to fix without success by adding setuptools-scm to build time dependencies here : https://github.com/pytest-dev/pytest-html/compare/6866776e86730ebbb2c10f02f089aeb84ddc4ba7..9986c6467015fcf8f36413189490535dfa90b858

Pierre-Sassoulas:
Something strange is happenning with dependabot see here : https://github.com/pytest-dev/pytest-html/pull/835. It seems we might need to upgrade manually.

Pierre-Sassoulas:
Maybe we can remove the build check while we reestablish the dependabot automated upgrading, then bring it back once everything is up to date. (Because manually upgrading is now that fun 😄 )

Pierre-Sassoulas:
All right there was two hynek used ! Fixed I think.

BeyondEvil (approved):
🚀

PR #977, opened 2025-11-23
Feat: Added a copy to clipboard feature for copying the test id.

Ambuj03:
As I suggested in discussions https://github.com/pytest-dev/pytest-html/discussions/976.
I have tried to add this little feature, Please give the feedback. 
Will be happy to make any required changes.

### DIFF
--- a/src/pytest_html/basereport.py
+++ b/src/pytest_html/basereport.py
@@ -284,7 +284,23 @@ class BaseReport:
         ]
         cells = [
             f'<td class="col-result">{outcome}</td>',
-            f'<td class="col-testId">{test_id}</td>',
+            f"""<td class="col-testId">{test_id}
+                <button class="copy-btn"
+                        data-test-id="{test_id}"
+                        title="Copy test ID"
+                        aria-label="Copy test ID">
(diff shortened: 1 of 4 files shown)


src/pytest_html/basereport.py
```diff
@@ -284,7 +284,23 @@ def _process_report(self, report, duration, processed_extras):
         ]
         cells = [
             f'<td class="col-result">{outcome}</td>',
-            f'<td class="col-testId">{test_id}</td>',
+            f"""<td class="col-testId">{test_id}
+                <button class="copy-btn"
+                        data-test-id="{test_id}"
+                        title="Copy test ID"
+                        aria-label="Copy test ID">
+                    <svg xmlns="http://www.w3.org/2000/svg"
```
RonnyPfannschmidt: how feasible is it to use U+1F4CB instead of the svg - i'd much prefer the character - but i have no idea how well supported that is

Ambuj03: Actually this unicode is better supported across all the browsers and mobile. Updated the PR with this change. 
Sample_CopytoClipboard

PR #1040, opened 2026-06-18
fix: make non-base64 media paths relative to report location

### LINKED ISSUE #309: links generated in the html file for attachments are absolute, there is no way to make them relative
Not sure if this is by design or if it is a bug, but if you create a report with attachments like PNG images, the html HREF for those images is the full path.

This has the adverse downside of making reports unusable if you send the report to anyone as zip file, or if you even move the folder where the report and related files are located.

To reproduce:

- Create a generic report, adding a png file
extra.append(pytest_html.extras.png("c:\foo\bar\img.png", name="my cool screen")
- Once you launch the html report, it will have the image displayed correctly.
- Move the html file, the resource folder and the png file somewhere else on your machine
- Open the HTML file again

Results:
- the html file will display a broken link for the image. The tag point at the original location so as expected, the image won't display unless the PNG file is in the link location.

decembercomposer697-hue:
Fixes #309: when extra media content is a file path (not base64),
it was used as-is in HTML, creating absolute paths. Now resolves
relative to the report's parent directory.

✅ Unit tests pass (7/7)

### DIFF
--- a/src/pytest_html/report.py
+++ b/src/pytest_html/report.py
@@ -32,8 +32,13 @@ class Report(BaseReport):
             media_data = base64.b64decode(content.encode("utf-8"), validate=True)
             return self._write_content(media_data, asset_name)
         except binascii.Error:
-            # if not base64 content, just return as it's a file or link
-            return content
+            # if not base64 content, try to make it relative to the report
+            try:
+                abs_path = Path(content).resolve()
+                return str(abs_path.relative_to(self._report_path.parent.resolve()))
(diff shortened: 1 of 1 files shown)


decembercomposer697-hue:
Hi! This PR fixes the issue where non-base64 media paths were absolute instead of relative to the report location. All unit tests pass. Would appreciate a review!

RonnyPfannschmidt (changes requested):
This needs some tests added that would fail without the change

